PR #1, opened 2019-03-14
Add initial end-to-end specs

stettix:
This PR adds some example code for writing data to the main table types (snapshot, single (date) partition, and multi-partition tables), and a set of specs that check that data can be written in a versioned manner.

These tests don't pass yet as they're meant to be high-level specs for the table versioning functionality. The idea is to use these to drive the implementation of the functionality that makes this work.

It is perhaps a bit unusual to merge tests that don't pass yet but I think it makes sense here, I'd rather merge to master frequently as the code evolves rather than have everything sitting in a long-lived branch.


build.sbt
```diff
@@ -17,48 +20,13 @@ lazy val commonSettings = Seq(
...
-  "-Ywarn-inaccessible", // Warn about inaccessible types in method signatures.
-  "-Ywarn-infer-any", // Warn when a type argument is inferred to be `Any`.
-  "-Ywarn-nullary-override", // Warn when non-nullary `def f()' overrides nullary `def f'.
-  "-Ywarn-nullary-unit", // Warn when nullary methods return Unit.
-  "-Ywarn-numeric-widen", // Warn when numerics are widened.
-  "-Ywarn-value-discard", // Warn when non-Unit expression results are unused.
-  "-language:higherKinds" // allow higher kinded types without `import scala.language.higherKinds`
+lazy val testSettings = Defaults.itSettings ++ Seq(
+  testOptions in Test += Tests.Argument(TestFrameworks.ScalaTest, "-oDF"),
+  parallelExecution in Test := false,
```
tomwadeson: What's the reason for not parallelising the unit test runs?

examples/src/main/scala/com/gu/tableversions/examples/AdImpressionLoader.scala
```diff
@@ -0,0 +1,63 @@
...
+import org.apache.spark.sql.{Dataset, SaveMode, SparkSession}
+
+/**
+  * This example contains code that writes example event data to a table with multiple partition columns.
+  * It demonstrates how individual partitions in such a table can be updated using versioning.
+  *
+  * @param tableName the fully qualified table name that will be populated by this loader
+  * @param tableUri The location where the table data will be stored
+  */
+class AdImpressionLoader(tableName: String, tableUri: String)(implicit val spark: SparkSession) {
```
tomwadeson: Wondering if naming these examples quite explicitly might be useful: `TableWithMultiplePartitions`?

examples/src/main/scala/com/gu/tableversions/examples/PageviewLoader.scala
```diff
@@ -0,0 +1,60 @@
...
+    """.stripMargin
+
+    spark.sql(ddl)
+    ()
+  }
+
+  def pageviews(): Dataset[Pageview] =
+    spark.table(tableName).as[Pageview]
+
+  def insert(dataset: Dataset[Pageview])(implicit spark: SparkSession): Unit = {
```
tomwadeson: Can we use the `SparkSession` from the constructor? Alternatively, can we pass the `SparkSession` into each method that requires it (implicitly or otherwise)

tomwadeson (approved):
Nice work! These tests are definitely serving their purpose in terms of documenting the use-cases - additional bonus that they're executable 😁

build.sbt
stettix: They use Spark and you can't run Spark tests in parallel

examples/src/main/scala/com/gu/tableversions/examples/PageviewLoader.scala
stettix: Ah yes, it should use the constructor one, that's a not-quite-complete refactoring.

examples/src/main/scala/com/gu/tableversions/examples/AdImpressionLoader.scala
stettix: Yeah I've been through a few iterations of names, it used to be something like that! 😊

But maybe `TableWithMultiplePartitionsLoader` would be better, I can change that.

build.sbt
stettix: As discussed IRL: I've move this from the common settings into the `examples` module which has these "end to end spec" type tests. You're right in that other modules probably want to run unit tests in parallel.

PR #2, opened 2019-03-21
Outline main model types and APIs

stettix:
This PR takes the previous 'loader' examples and changes them to write data in a versioned way, instead of the plain old Spark/Hive way as they did before. The main aim for this is to discover the key APIs that we need for these use cases, and the related model types.

This also includes a basic (but inefficient) implementation of a versioned 'insert' operation on Spark `Datasets` that writes to the versioned storage layout, i.e. writes to `s3://bucket/data/v42` etc for snapshot tables and to `s3://bucket/data/date=2019-01-31/v42` etc for partitioned tables.

What this PR does _not_ implement, and which is left for later PRs, is providing implementations for the `Metastore` and `TableVersion` APIs - we just have empty placeholders for these. Also, the multi-partitioned example hasn't been fleshed out yet.


core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -0,0 +1,53 @@
...
+    * Start tracking version information for given table.
+    * This must be called before any other operations can be performed on this table.
+    */
+  def init(table: TableName): F[Unit]
+
+  /** Get details about partition versions in a table. */
+  def currentVersion(table: TableName): F[TableVersion]
+
+  /** Get a description of which version to write to next for the given partitions of a table. */
+  def nextVersions(table: TableName, partitions: List[Partition]): F[List[PartitionVersion]]
```
desbo, line 22: Why does this return multiple versions?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -0,0 +1,16 @@
+package com.gu.tableversions.core
+
+/**
+  * Reference implementation of the table version store. Does not persist state.
+  */
+class InMemoryTableVersions[F[_]] extends TableVersions[F] {
```
desbo, line 6: Will this be used outside of tests? If not, I'd probably put it under `test`

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -1,8 +1,73 @@
...
+final case class PartitionVersion(partition: Partition, version: VersionNumber)
+
 //
-// A partition schema describes the fields used for partitions of a table
+// Tables
 //
-case class PartitionColumn(name: String) extends AnyVal
 
-case class PartitionSchema(keys: List[PartitionColumn])
+final case class TableName(schema: String, name: String) {
```
desbo, line 64: Should we use the one from `tableMetadata`?

stettix: `tableMetadata` lives in the ophan-data-lake project, and we don't want this project to depend on that.

examples/src/main/scala/com/gu/tableversions/examples/SnapshotTableLoader.scala
```diff
@@ -1,46 +1,54 @@
...
+    spark.table(table.name.fullyQualifiedName).as[User]
+
+  def insert(dataset: Dataset[User], userId: UserId, message: String): Unit = {
+    import VersionedDataset._
+
+    val (latestVersion, metastoreChanges) = dataset.versionedInsertInto(table, userId, message)
+
+    logger.info(s"Updated table $table, new version details:\n$latestVersion")
+    logger.info(s"Applied the the following changes to sync the Metastore:\n$metastoreChanges")
+  }
```
stettix, line 51: These 'loader' classes are becoming so similar now across snapshot + partitioned cases that they could perhaps be unified. Haven't really prioritised it though, as it's example / test code.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -0,0 +1,53 @@
...
+/**
+  * This defines the interface for querying and updating table version information tracked by the system.
+  */
+trait TableVersions[F[_]] {
+
+  /**
+    * Start tracking version information for given table.
+    * This must be called before any other operations can be performed on this table.
+    */
+  def init(table: TableName): F[Unit]
```
tomwadeson, line 16: We could enforce that this is called before other operations can be performed by returning some sort of token

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
tomwadeson, line 22: `F[Map[Partition, PartitionVersion]]` might be a bit more convenient

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -0,0 +1,53 @@
...
+      timestamp: Instant,
+      updatedPartitions: List[PartitionVersion])
+
+  case class UpdateMessage(content: String)
+
+  case class UserId(value: String)
+
+  /** Result type for commit operation */
+  sealed trait CommitResult
+  object SuccessfulCommit extends CommitResult
```
tomwadeson: `case object`?

And I'm quite a fan of namespacing the constructors within a companion of the `trait`.

dynamodb/src/main/scala/com/gu/tableversions/dynamodb/DynamoDbTableStore.scala
```diff
@@ -1,9 +1,18 @@
 package com.gu.tableversions.dynamodb
 
-import com.gu.tableversions.core.TableStore
+import com.gu.tableversions.core._
 
-class DynamoDbTableStore[F[_]] extends TableStore[F] {
+/**
+  * Placeholder for DynamoDB implementation of the table versions service.
+  */
+class DynamoDbTableStore[F[_]] extends TableVersions[F] {
```
tomwadeson: I wonder if this PR could be a bit more cohesive without the unused & unimplemented services?

examples/src/main/scala/com/gu/tableversions/examples/DatePartitionedTableLoader.scala
```diff
@@ -1,47 +1,56 @@
...
 /**
   * This example contains code that writes example event data to a table that has a single date partition.
   * It demonstrates how individual partitions in such a table can be updated using versioning.
-  *
-  * @param tableName the fully qualified table name that will be populated by this loader
-  * @param tableLocation The location where the table data will be stored
   */
-class DatePartitionedTableLoader(tableName: String, tableLocation: URI)(implicit val spark: SparkSession) {
+class DatePartitionedTableLoader(table: TableDefinition)(
+    implicit tableVersions: TableVersions[IO],
```
tomwadeson, line 19: I think it'd be simpler to pass these services explicitly

metastore/src/main/scala/com/gu/tableversions/metastore/Metastore.scala
```diff
@@ -0,0 +1,46 @@
...
+import com.gu.tableversions.metastore.Metastore.TableChanges
+
+trait Metastore[F[_]] {
+
+  /**
+    * Describe the current table in the Metastore interpreted in terms of version information.
+    *
+    * @param table The table to query
+    */
+  def currentVersion(table: TableName): F[TableVersion]
```
tomwadeson: `F[Option[TableVersion]]`?

metastore/src/main/scala/com/gu/tableversions/metastore/Metastore.scala
```diff
@@ -0,0 +1,46 @@
...
+    Metastore.computeChanges(current, target)
+
+}
+
+object Metastore {
+
+  final case class TableChanges(operations: List[TableOperation])
+
+  sealed trait TableOperation
+  final case class AddPartition(partition: PartitionVersion) extends TableOperation
```
tomwadeson: Companion namespace thing, if you agree

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -1,3 +1,130 @@
...
+import com.gu.tableversions.metastore.{Metastore, VersionPaths}
+import org.apache.spark.sql.{Dataset, Row, SparkSession}
+
+/**
+  * Code for writing Spark datasets to storage in a version-aware manner, taking in version information,
+  * using the appropriate paths for storage, and committing version changes.
+  */
+object VersionedDataset {
+
+  implicit def toDatasetOps[T](
```
tomwadeson: `implicit class` captures this pattern - I think @desbo mentioned to me that it might have issues in some instances, but I don't recall the details. I like it anyway - a bit less mechanical!

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -1,3 +1,130 @@
...
+import com.gu.tableversions.core._
+import com.gu.tableversions.metastore.Metastore.TableChanges
+import com.gu.tableversions.metastore.{Metastore, VersionPaths}
+import org.apache.spark.sql.{Dataset, Row, SparkSession}
+
+/**
+  * Code for writing Spark datasets to storage in a version-aware manner, taking in version information,
+  * using the appropriate paths for storage, and committing version changes.
+  */
+object VersionedDataset {
```
tomwadeson, line 17: I think this should be written against some `F[_]` with the required `cats-effect` constraints, otherwise we impose `IO` on users.

spark/src/test/scala/com/gu/tableversions/spark/VersionedDatasetSpec.scala
```diff
@@ -1,15 +1,266 @@
...
+    val nextPartitionVersions = List(PartitionVersion(Partition.snapshotPartition, VersionNumber(1)))
+
+    val stubbedChanges = TableChanges(nextPartitionVersions.map(AddPartition))
+    implicit val stubMetastore: Metastore[IO] = new StubMetastore(
+      currentVersion = initialTableVersion,
+      computedChanges = stubbedChanges
+    )
+
+    // Stub table versions
+    val committedTableUpdatesRef = Ref[IO].of(List.empty[TableVersions.TableUpdate]).unsafeRunSync()
```
tomwadeson, line 122: I think this could be reformulated a tad to just have the one `unsafeRunSync`

Not for some ideological reason, but because I think it'll make the interactions between the various pieces a bit snappier.

tomwadeson (commented):
Had a quick scan and think it looks really good. I'll take a look in a bit more detail tomorrow.

core/src/main/scala/com/gu/tableversions/core/model.scala
desbo: duhhh, of course 😳

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: It's a version per partition. So if you're inserting a dataset that contains three days worth of data, then we need to get the appropriate version for each date partition that this will be written to.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Interesting point. My initial thought was that this only needs to be called once per table, e.g. the first time the table is populated (a bit like you call `git init` when you first initialise a git repo). But as we want jobs to auto-create new tables, we'd either need an `exists` method - or just make the `init` method idempotent so it can be called every time the job runs.

But! I can imagine use cases, e.g. in the command line for this, where you may want to call one of the other operations on the version store, where you definitely wouldn't want to call `init` first. E.g. if you run a command like `table-versions log clean.foobarbaz` you would probably want this to return an error saying `No table 'clean.foobarbaz' found`, rather than going and creating it in the version store.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Whoops yes!

dynamodb/src/main/scala/com/gu/tableversions/dynamodb/DynamoDbTableStore.scala
stettix: Sure, these were just placeholders to show intent, I can remove them if that's clearer.

examples/src/main/scala/com/gu/tableversions/examples/DatePartitionedTableLoader.scala
stettix: The reason I made them implicit here is that they are implicit arguments in the insert operation on the `Dataset`. Would you want to pass them explicitly there as well?

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
stettix: Great, I'll do that, I'm not sure about the preferred way to do this these days!

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
stettix: Yes agreed, I just did `IO` here initially to get something out the door faster. Do you think it's worth fixing this now, later in the PoC, or leave for potential later iterations?

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
tomwadeson: You've just made me wonder if `git init` is idempotent 😄 

Yeah, I think some something like a:

```scala
def init(name: String): F[Table.Id]
def exists(name: String): F[Option[Table.Id]]
```

pair would be nice. Then all other ops pass this `Table.Id` around.

spark/src/test/scala/com/gu/tableversions/spark/VersionedDatasetSpec.scala
stettix: Yes, I felt dirty writing this, and had an apology in a comment at some point 😊. This is test code, which is why I reckoned it was just about alright, but maybe I should tidy it up.

(What do you mean by "snappier" in this context by the way?)

examples/src/main/scala/com/gu/tableversions/examples/DatePartitionedTableLoader.scala
tomwadeson: I see! My gut says "yes", but maybe it's too early to say. Something to ponder :)

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Not sure if having an `Id` type adds much in that case, as the `TableName` already is the unique ID for the table? Happy to discuss anyway!

spark/src/test/scala/com/gu/tableversions/spark/VersionedDatasetSpec.scala
tomwadeson: Yeah, "snappier" is pretty sloppy, but I mean "snaps together nicely". So... compositional!

tomwadeson: I just think there are big wins to be had from writing in this style, and it's cool to realise those wins in test code just as much (OK... almost as much!) as main.

I take your earlier point though that this is in POC status right now, however. I think that's totally fair and would colour anything I'd review, so I think in that sense it might be good to hear what bits you think are out-of-scope from a review POV.

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
tomwadeson: That's fair! I don't think I have an opinion then - whatever you think meets the requirements of this POC.

stettix: How about I pair with you at some point later in this sprint to have a look at this particular change? I'd need some input on how to best do this but I'd prefer it to not block this PR.

stettix:
@tomwadeson Regarding your question about what's worth commenting on/fixing in this PR given this is a PoC, and given that this PR is only one step towards getting the initial end-to-end scenarios working:

I think it's a question of striking a balance - I don't want to compromise on code quality as I think that's a false economy. On the other hand, the main purpose of this sprint's PoC is to prove the central ideas of the table versioning proposal and clarify if/how this would work. If we decide to go forward with this then I think some productionising & hardening of the solution would need to happen.

So how about considering these areas as things to leave for later:

- Detailed handling of error scenarios, detailed return types for errors, helpful error messages, comprehensive logging
- Performance
- Genericising the solution (e.g. for other effect types than `IO`, but I'd say supporting other Scala and Spark versions would fall into the same category)
- Packaging and publishing

Does that make sense at all?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: Thinking about it, it might be useful to have a stub version of this for trying running this code in an EMR cluster against a real metastore + S3... so I might leave it place until we have a real implementation?

core/src/main/scala/com/gu/tableversions/core/model.scala
stettix: Also, I had a look at that to see if I could, ahem, borrow the code for it. But, that's a bit different; the storage path for the table lives in the same class as the name and schema (in `TableReference`), and the schema is fixed to a hard-coded list of schema names, so it doesn't really fit here.

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
tomwadeson: Sounds great!

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: I tried this, but didn't help simplify the calling code. Also, that type signature seems slightly looser as in the type wouldn't enforce that a `Partition` key in the `Map` refers to a `PartitionVersion` value that contains the same `Partition`. So I'm inclined to stick with it as it is.

examples/src/main/scala/com/gu/tableversions/examples/DatePartitionedTableLoader.scala
```diff
@@ -1,47 +1,56 @@
...
-    // with new functionality in this project.
-    dataset.write
-      .mode(SaveMode.Overwrite)
-      .insertInto(tableName)
+    spark.table(table.name.fullyQualifiedName).as[Pageview]
+
+  def insert(dataset: Dataset[Pageview], userId: UserId, message: String): Unit = {
+    import VersionedDataset._
+
+    val (latestVersion, metastoreChanges) = dataset.versionedInsertInto(table, userId, message)
```
stettix, line 50: This is the key bit of this PR, in that this is the code that uses the new API for inserting a dataset into a table in a versioned way as opposed to the previous "insert via Hive" way.

spark/src/test/scala/com/gu/tableversions/spark/VersionedDatasetSpec.scala
stettix: See comment elsewhere for a discussion about what to include in the PoC. About the specific suggested fix:

I had a quick go at this but I think it works out quite unnaturally in this particular test. Basically, I think the APIs in the `spark` module should follow Spark conventions hence throw exceptions. I.e. I don't think it makes sense to have operations on a `Dataset[T]` that returns `IO[U]`. The reason `IO` comes up here is just for setting up the stubs that this class interacts with - it's not used by the actual code under test. In which case, I'm not convinced it's worth changing this. Or am I missing something?

tomwadeson (approved):
Nice work!

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
JonNorman: I took it as the only way of getting a `Table.Id` is through calling one of `init` or `exists` and thus this ensures that one has been called before doing the other dependent operations...?

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
JonNorman: So the delta between contiguous versions is only ever going to be a single partition?

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Yes, I think that was the idea, but my concern is that there's no guarantee that the table is still there by the time a client calls one of the other operations. I.e. some other client may have deleted the table. So, you have to deal with those cases anyway, in which case I don't know if the extra level of indirection adds much.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Not sure if I understand the question, but a single update to a table can contain a number of updated partitions. Does that make sense?

JonNorman: I think I've misread the cardinality:

> It's a version per partition.

I took that to mean that we get a new version for each updated partition.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -0,0 +1,56 @@
...
+  /** A collection of updates to partitions to be applied and tracked as a single atomic change. */
+  final case class TableUpdate(
+      userId: UserId,
+      message: UpdateMessage,
+      timestamp: Instant,
+      updatedPartitions: List[PartitionVersion])
+
+  final case class UpdateMessage(content: String)
+
+  final case class UserId(value: String)
```
JonNorman: Nitpick, but `extends AnyVal`?

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -1,8 +1,73 @@
...
+        columnValues.map(columnValue => s"${columnValue.column.name}=${columnValue.value}").mkString("", "/", "/")
+      normalised(tableLocation).resolve(partitionsSuffix)
+    }
+  }
+}
+
+object Partition {
+
+  /** Convenience constructor for single column partitions. */
+  def apply(columnValue: ColumnValue): Partition = Partition(List(columnValue))
```
JonNorman, line 26: Any reason not to make this:

```scala
def apply(columnValues: ColumnValue*): Partition = Partition(columnValues.toList)
```
?

examples/src/main/scala/com/gu/tableversions/examples/SnapshotTableLoader.scala
JonNorman: I guess this makes sense since we are treating snapshots as a special case of partitioned data sets. Looks good!

examples/src/test/scala/com/gu/tableversions/examples/DatePartitionedTableLoaderSpec.scala
```diff
@@ -25,7 +39,8 @@ class DatePartitionedTableLoaderSpec extends FlatSpec with Matchers with SparkHi
       Pageview("user-3", "sport/football", Timestamp.valueOf("2019-03-13 21:00:00"))
     )
 
-    loader.insert(pageviewsDay1.toDS().coalesce(2))
+    val userId = UserId("test user")
+    loader.insert(pageviewsDay1.toDS().coalesce(2), userId, "Day 1 initial commit")
```
JonNorman: Why are you using `coalesce` here?

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
```diff
@@ -0,0 +1,35 @@
...
+object VersionPaths {
+
+  /**
+    * @return the fully resolved path for collection of versioned path
+    */
+  def resolveVersionedPartitionPaths(
+      partitionVersions: List[PartitionVersion],
+      tableLocation: URI): Map[Partition, URI] = {
+
+    partitionVersions.map { partitionVersion =>
```
JonNorman, line 19: I know this is just a reference implementation at this point, but (nitpick): would it be tidier to have
```scala
def resolveVersionedPartitionPath(partitionVersion: PartitionVersion, tableLocation: URI): URI = ???
```

which is then called by `resolveVersionedPartitionPaths`?

metastore/src/test/scala/com/gu/tableversions/metastore/VersionPathsSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import java.net.URI
+
+import com.gu.tableversions.core.Partition.PartitionColumn
+import com.gu.tableversions.core.{Partition, PartitionVersion, VersionNumber}
+import org.scalatest.{FlatSpec, Matchers}
+
+class VersionPathsSpec extends FlatSpec with Matchers {
+
+  "Resolving a path with version 0" should "produce the original path" in {
```
JonNorman, line 11: Is this behaviour to capture the use case we discussed for making migration easier i.e. not moving all of our data into `v0` prefixes?

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -1,3 +1,131 @@
...
+      datasetPartitions <- IO(VersionedDataset.partitionValues(dataset, table.partitionSchema)(dataset.sparkSession))
+
+      // Get next version numbers for the partitions of the dataset
+      workingVersions <- tableVersions.nextVersions(table.name, datasetPartitions)
+
+      // Resolve the path that each partition should be written to, based on their version
+      partitionPaths = VersionPaths.resolveVersionedPartitionPaths(workingVersions, table.location)
+
+      // Write Spark dataset to the versioned path
+      _ <- IO(VersionedDataset.writeVersionedPartitions(dataset, partitionPaths))
```
JonNorman, line 57: Do we want to try and do any cleanup if this succeeds but updating the metastore doesn't work? I suppose we probably wouldn't have to since the next time we try and write, the new (hidden) versioned data will be overwritten...?

examples/src/test/scala/com/gu/tableversions/examples/DatePartitionedTableLoaderSpec.scala
stettix: I just wanted to limit the number of partitions in the tests data. It might be pointless, I'll try removing it.

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
stettix: I guess that's what `pathFor` is - I was going to make that private, but maybe I should rename it to `resolveVersionedPartitionPath` instead?

metastore/src/test/scala/com/gu/tableversions/metastore/VersionPathsSpec.scala
stettix: That's the intention, yes.

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
stettix: Yes the latter, I'd say. Cleaning up is just more stuff to do that might fail, and I'm not sure it's useful.

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
JonNorman: Yeah, I don't feel strongly either way, just raising it as something to consider. Looks good as it is 👍

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -0,0 +1,56 @@
+package com.gu.tableversions.core
+
+import java.time.Instant
+
+import com.gu.tableversions.core.TableVersions.CommitResult
+
+/**
+  * This defines the interface for querying and updating table version information tracked by the system.
+  */
+trait TableVersions[F[_]] {
```
desbo, line 10: I'm thinking `rollback(partition: List[Partition])` (to set the provided partitions to the previous version) will be a pretty widely used feature. Do you think it should be a part of this core API?

stettix: Absolutely! But I didn't write out the specs for that scenario in this PR as I just aimed to get the basic scenarios working first.

core/src/main/scala/com/gu/tableversions/core/model.scala
desbo: I'm guessing it doesn't make sense to call that with an empty parameter list?

stettix: Ah yes, good point!

JonNorman: Don't we do that here? https://github.com/guardian/table-versions/pull/2/files/1f7ee1f4992013e443e29c76ae31d44a0a5d68c6#diff-233f9d1b6e81ff49c1ced307784690ccR36

TBonnin:
Great job @stettix 
As mentioned face to face I am only questioning the concept of asking for the next version before to write the data. It conflicts with the concept of truly immutable files. Things can happen between the `nextVersion` and the `commit` operations.
I would propose writing to a location defined by convention (ex: timestamp-UDID) which would be truly unique for every single write. It is a bit less human readable but you can offset the issue by adding a metadata file in the subfolder with the version number during commit (similar to the spark SUCCESSFULL file)

stettix:
> I would propose writing to a location defined by convention (ex: timestamp-UDID) which would be truly unique for every single write. It is a bit less human readable but you can offset the issue by adding a metadata file in the subfolder with the version number during commit (similar to the spark SUCCESSFULL file)

I like that idea and definitely think it's worth exploring. I'd be tempted to get this PR in first though so other work can proceed, then try changing this in a later PR at some point, if that's OK with you.

TBonnin:
No problem. We can iterate and improve in following PRs

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -1,8 +1,73 @@
...
+
+}
+
+//
+// Versions
+//
+
+final case class VersionNumber(number: Int) extends AnyVal
+
+final case class PartitionVersion(partition: Partition, version: VersionNumber)
```
TBonnin, line 58: Should the PartitionVersion have a `creationDate` field? Might be useful to know when it was created

stettix: Sounds useful, but this type is used in the response from the metastore as well and I don't know if I can produce this information from there. So it might fit in a different type.

I suspect this and a number of similar points will be highlighted once we implement scenarios that include showing the version history and doing rollback to older versions. The scenarios in this PR wouldn't make use of this information.

core/src/main/scala/com/gu/tableversions/core/model.scala
stettix: That's a good point too! 

So I tried this, but if I make it varargs, it conflicts with the other `apply` method. I'll leave it as it is for now, expecting to review the model API later. (I find it a bit verbose as it is, it would be nice if we could fix that as well)

PR #3, opened 2019-03-26
Implement Spark/Hive Metastore service

stettix:
This PR adds an implementation of the `Metastore` trait that is backed by the Spark/Hive metastore, accessed via Spark SQL. Note that we don't expect this one to work with the AWS Glue metastore; we will probably need a separate implementation for that. But we reckon it's still useful to have this implementation - mainly to be able to run end-to-end tests locally, but also for anyone who's actually using a Hive Metastore instead of Glue.

Also included are some generic tests that can be applied to any future implementations of the `Metastore` API.


metastore/src/main/scala/com/gu/tableversions/metastore/Metastore.scala
```diff
@@ -11,14 +11,16 @@ trait Metastore[F[_]] {
     * Describe the current table in the Metastore interpreted in terms of version information.
     *
     * @param table The table to query
+    * @return the table version information, or a failure if the table doesn't exist or can't be queried.
     */
-  def currentVersion(table: TableName): F[Option[TableVersion]]
+  def currentVersion(table: TableName): F[TableVersion]
```
stettix, line 20: I changed the return type for this back to `F[TableVersion]` as I found that a) it was confusing what `None` meant, and b) all caller code wanted to treat this as an error anyway so had to explicitly turn this into a failure.

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  override def update(table: TableName, changes: Metastore.TableChanges): F[Unit] =
+    changes.operations.map(appliedOp(table)).sequence.void
+
+  private def appliedOp(table: TableName)(operation: TableOperation): F[Unit] =
+    operation match {
+      case AddPartition(partitionVersion)                    => addPartition(table, partitionVersion)
+      case UpdatePartitionVersion(partitionVersion)          => updatePartitionVersion(table, partitionVersion)
+      case RemovePartition(partition)                        => removePartition(table, partition)
+      case UpdateTableLocation(tableLocation, versionNumber) => updateTableLocation(table, tableLocation, versionNumber)
```
TBonnin: I find it confusing that the `UpdateTableLocation` operation (which is the only operation acting on the table level) is taking a version number. Isn't the concept of version only applying to a partition, not a table?

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
```diff
@@ -0,0 +1,183 @@
...
+
+  private[spark] def toPartitionExpr(partitionPath: String): String =
+    toHivePartitionExpr(parsePartition(partitionPath))
+
+  private[spark] def toHivePartitionExpr(partition: Partition): String =
+    partition.columnValues
+      .map(columnValue => s"${columnValue.column.name}='${columnValue.value}'")
+      .mkString("(", ",", ")")
+
+  private val ColumnValueRegex = "([a-z_]+)=(.+)".r
```
desbo: Minor, but I like the `x` flag to allow comments in regexes longer than more than a few characters. For example, this could be written as: 

```suggestion
  private val ColumnValueRegex = """
      |(?x)
      |([a-z_]+)  # column name
      |=
      |(.+)       # column value
    """.stripMargin
```

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+
+    it should "allow table versions to be updated for snapshot tables" in {
+
+      val scenario = for {
+        metastore <- emptyMetastore
+        _ <- initHiveTable
+
+        initialVersion <- metastore.currentVersion(table.name)
+
+        _ <- metastore.update(table.name, TableChanges(List(UpdateTableLocation(table.location, VersionNumber(1)))))
```
TBonnin: I don't think the version number should be manipulated explicitely by the end user. However do we want to abstract it within `Metastore` or at a higher level?

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
stettix: TIL!

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
stettix: In the case of snapshot tables, it is the main table location we have to change and not partition versions. So the question is how we model that through this API. I guess the other option is to use `UpdatePartitionVersion` for this as well, using the 'special' partition for snapshots. This may well be better, I'll try to do this and see how it looks.

Actually, that reminds me: I'm considering updating the model to not use the "special" / empty partition for snapshot tables but have ADTs with explicit types for the partitioned and versioned cases instead. That might play into how this looks too.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
stettix: See dicsussion ~~above~~ below, but I don't think this interface is something for the "end user", it's a fairly low level service the makes an underlying metastore version aware (hmm, maybe `VersionAwareMetastore` or `VersionedMetastore` would be a better name for this).

In the real code, the versions passed into the metastore's methods would come from the version store (`TableVersions` right now) and not be magicked up by clients..

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
```diff
@@ -0,0 +1,182 @@
...
+class SparkHiveMetastore[F[_]](implicit spark: SparkSession, F: Sync[F]) extends Metastore[F] with LazyLogging {
+
+  import SparkHiveMetastore._
+  import spark.implicits._
+
+  override def currentVersion(table: TableName): F[TableVersion] = {
+
+    val partitionedTableVersion: F[TableVersion] = for {
+      partitions <- listPartitions(table)
+      partitionLocations <- partitions.map { partition =>
```
tomwadeson: A couple of opportunities for `traverse` here and below in `def update`:

```scala
.map(someApplicativeAction).sequence === .traverse(someApplicativeAction)
```

In the case of `def update`, `traverse_` (note the underscore) captures the `.void` bit too: "I don't care about the result, just sequence all of these actions for their effects"

I've also started to find that when I'm reifying control flow with `IO` and `traverse` and `flatMap`, it's a signal that pushes me towards fs2. We're perhaps not there yet, but maybe it's not too far off! 😁

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
```diff
@@ -0,0 +1,175 @@
...
+import com.gu.tableversions.metastore.Metastore.TableChanges
+import com.gu.tableversions.metastore.Metastore.TableOperation._
+import org.scalatest.{FlatSpec, Matchers}
+
+/**
+  * Spec containing tests that apply across all Metastore implementations.
+  *
+  * These are black box tests purely in terms of the Metastore interface.
+  */
+trait MetastoreSpec {
```
tomwadeson, line 15: These tests are lovely!

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
```diff
@@ -0,0 +1,182 @@
...
+    } yield TableVersion(List(PartitionVersion(Partition.snapshotPartition, versionNumber)))
+
+    // Choose the calculation to perform based on whether we have a partitioned table or not
+    isPartitioned(table).flatMap(partitioned => if (partitioned) partitionedTableVersion else snapshotTableVersion)
+  }
+
+  override def update(table: TableName, changes: Metastore.TableChanges): F[Unit] =
+    changes.operations.map(appliedOp(table)).sequence.void
+
+  private def appliedOp(table: TableName)(operation: TableOperation): F[Unit] =
```
tomwadeson, line 48: Lovely lovely lovely

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
```diff
@@ -0,0 +1,182 @@
...
+  private def removePartition(table: TableName, partition: Partition): F[Unit] = {
+    val partitionExpr = toHivePartitionExpr(partition)
+    val removePartitionQuery =
+      s"ALTER TABLE ${table.fullyQualifiedName} DROP IF EXISTS PARTITION $partitionExpr"
+    performUpdate(s"Removing partition $partition from table ${table.fullyQualifiedName}", removePartitionQuery)
+  }
+
+  private def versionedPartitionLocation(table: TableName, partitionVersion: PartitionVersion): F[URI] =
+    for {
+      tableLocation <- findTableLocation(table).map(new URI(_))
```
tomwadeson: I've not checked other usages, but might this always be a `URI`?

spark/src/test/scala/com/gu/tableversions/spark/SparkHiveMetastoreSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+import cats.effect.IO
+import cats.syntax.functor._
+import com.gu.tableversions.core.Partition.{ColumnValue, PartitionColumn}
+import com.gu.tableversions.core._
+import com.gu.tableversions.metastore.MetastoreSpec
+import org.scalatest.prop.PropertyChecks
+import org.scalatest.{FlatSpec, Matchers}
+import SparkHiveMetastore._
+
+class SparkHiveMetastoreSpec extends FlatSpec with Matchers with SparkHiveSuite with MetastoreSpec with PropertyChecks {
```
tomwadeson, line 14: 👍

spark/src/test/scala/com/gu/tableversions/spark/SparkHiveMetastoreSpec.scala
```diff
@@ -0,0 +1,138 @@
...
+        //"/",
+        "")
+    // format: on
+
+    forAll(invalidPartitionStrings) { partitionStr =>
+      an[Exception] should be thrownBy SparkHiveMetastore.parsePartition(partitionStr)
+    }
+  }
+
+  "Rendering a partition as a string" should "produce a valid Hive partition expression" in {
```
tomwadeson: Even if we don't use ScalaCheck, there's a round-trip property here that might be worth testing explicitly.

Maybe a single-column `Table` containing Hive partition expressions that when parsed and then serialised is unchanged.

tomwadeson (approved):
Nice work!

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
stettix: > A couple of opportunities for `traverse` here and below in `def update`
[...[
> In the case of `def update`, `traverse_` (note the underscore) captures the `.void` bit too: "I don't care about the result, just sequence all of these actions for their effects"

Ahhhh nice thanks! 🙌

> I've also started to find that when I'm reifying control flow with `IO` and `traverse` and `flatMap`, it's a signal that pushes me towards fs2. We're perhaps not there yet, but maybe it's not too far off! 😁

Yes, I did find myself missing fs2 while writing this.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
stettix: I had a look at this but on reflection, I think it's better that this particular interface is explicit about the changes it will perform to tables in the underlying metastore. I don't think this should leak into the higher level interfaces anyway. Might see if I can write some docs around this to explain what this is the case.

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
stettix: I couldn't find a spec that explicitly says so, but from looking around I believe this is the case.

That said, it would be better if the `URI` values were produced at origin and not passed around as `String`s, so I'll fix that.

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
stettix: Got it to work with some minor adjustments (the `(?x)` has to be before any newlines, and there's the `.r`):

```scala
  private val ColumnValueRegex = """(?x)
                                   |([a-z_]+)  # column name
                                   |=
                                   |(.+)       # column value
                                 """.stripMargin.r
```

spark/src/test/scala/com/gu/tableversions/spark/SparkHiveMetastoreSpec.scala
stettix: Roundtripping would be nice, but we don't have any code for parsing Hive partition expressions, only partition paths.

The former is expressions like `(event_date='2019-01-30', processed_date='2019-01-31')`, which is the syntax used in queries for adding partitions. I.e. we only render such strings for use in queries, we never parse them.

The latter is strings like `event_date=2019-01-30/processed_date=2019-01-31`, which is what we get back from "show partitions" queries. These ones we only parse when they come back, we don't produce them.

spark/src/main/scala/com/gu/tableversions/spark/SparkHiveMetastore.scala
stettix: As mentioned in a reply below, I think it's worth keeping the explicit operation to update a base table location, distinct from updating partitions. I did make a different change in this area though: I realised it didn't make sense to pass in the table location when updating the table version, so instead of:

```scala
case UpdateTableLocation(tableLocation, versionNumber) => updateTableLocation(table, tableLocation, versionNumber)
```

It's now:

```scala
case UpdateTableVersion(versionNumber)        => updateTableLocation(table, versionNumber)
```

Let me know what you think.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
TBonnin: Make sense. So this `Metastore` API (or `VersionedMetastore` 😉) is low level and will be used by the user-facing layer (which I guess will take care of keeping track of version history)

metastore/src/main/scala/com/gu/tableversions/metastore/Metastore.scala
```diff
@@ -35,15 +41,54 @@ object Metastore {
...
   }
 
-  def computeChanges(current: TableVersion, target: TableVersion): TableChanges = ???
+  def computeChanges(oldVersion: TableVersion, newVersion: TableVersion): TableChanges = {
+
+    val operations = if (oldVersion.partitionVersions.map(_.partition) == List(Partition.snapshotPartition)) {
+      assert(newVersion.partitionVersions.map(_.partition) == List(Partition.snapshotPartition),
+             "Can't change table from snapshot table to partitioned")
+
+      if (oldVersion.partitionVersions.head.version != newVersion.partitionVersions.head.version)
```
TBonnin, line 63: Is partitionVersions always garantueed to have a head?

TBonnin (approved):
Very nice work

metastore/src/main/scala/com/gu/tableversions/metastore/Metastore.scala
stettix: Yes, due to the assertion above.

That said, I don't really like how the code has worked out, treating snapshot tables as a special case which has a single empty version. I would like to try remodelling this explicitly, at least in most of the APIs, so we avoid these kinds of checks. For example making `TableVersion` an ADT with concrete cases for snapshot tables and partitioned tables.

I think that's for a later PR though.

TBonnin: I agree, making explicit and impossible to compile would be best.

PR #4, opened 2019-03-29
Implement in-memory version service

stettix:
This PR includes code for an in-memory implementation of the `TableVersions` trait, which is the service that tracks the history of table versions.

This makes the end-to-end specs pass 🎉 (with the odd bug fix), so they are enabled now.

### DIFF
--- a/core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
+++ b/core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
@@ -6,6 +6,7 @@ import cats.implicits._
 import com.gu.tableversions.core.InMemoryTableVersions.TableUpdates
 import com.gu.tableversions.core.TableVersions.CommitResult.SuccessfulCommit
 import com.gu.tableversions.core.TableVersions._
+import com.gu.tableversions.core.util.RichRef._
 
 /**
   * Reference implementation of the table version store. Does not persist state.
@@ -56,22 +57,12 @@ class InMemoryTableVersions[F[_]] private (allUpdates: Ref[F, TableUpdates])(imp
         Left(new Exception(s"Unknown table '${table.fullyQualifiedName}'"))
     }
 
-    InMemoryTableVersions.modifyEither(allUpdates)(applyUpdate).map(_ => SuccessfulCommit)
+    allUpdates.modifyEither(applyUpdate).map(_ => SuccessfulCommit)
   }
 
 }
 
 object InMemoryTableVersions {
-
-  // TODO: Put somewhere else, as syntax on `Ref`
-  def modifyEither[F[_]: Sync, A, E <: Throwable](ref: Ref[F, A])(f: A => Either[E, A]): F[Unit] =
-    ref
-      .modify(a =>
-        f(a) match {
-          case Left(e)     => (a, e.raiseError[F, Unit])
-          case Right(newA) => (newA, ().pure[F])
-      })
-      .flatten
 
   type TableUpdates = Map[TableName, List[TableUpdate]]
 
--- a/core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
+++ b/core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
@@ -1,16 +1,117 @@
 package com.gu.tableversions.core
 
+import cats.effect.Sync
+import cats.effect.concurrent.Ref
+import cats.implicits._
+import com.gu.tableversions.core.InMemoryTableVersions.TableUpdates
+import com.gu.tableversions.core.TableVersions.CommitResult.SuccessfulCommit
+import com.gu.tableversions.core.TableVersions.{AddPartitionVersion, PartitionOperation, RemovePartition, TableUpdate}
+
 /**
   * Reference implementation of the table version store. Does not persist state.
   */
-class InMemoryTableVersions[F[_]] extends TableVersions[F] {
+class InMemoryTableVersions[F[_]] private (allUpdates: Ref[F, TableUpdates])(implicit F: Sync[F])
+    extends TableVersions[F] {
 
-  override def init(table: TableName): F[Unit] = ???
+  override def init(table: TableName): F[Unit] =
+    allUpdates.update { prev =>
+      if (prev.contains(table)) prev else prev + (table -> Nil)
+    }
 
-  override def currentVersion(table: TableName): F[TableVersion] = ???
+  override def currentVersion(table: TableName): F[TableVersion] =
+    // Derive current version of a table by folding over the history of changes
+    for {
+      allTableUpdates <- allUpdates.get
+      tableUpdates <- allTableUpdates
+        .get(table)
+        .fold(F.raiseError[List[TableUpdate]](new Exception(s"Table '$table' not found")))(F.pure)
+      operations = tableUpdates.flatMap(_.partitionUpdates)
+    } yield InMemoryTableVersions.applyUpdate(TableVersion.empty)(operations)
 
-  override def nextVersions(table: TableName, partitionColumns: List[Partition]): F[List[PartitionVersion]] = ???
+  override def nextVersions(table: TableName, partitions: List[Partition]): F[List[PartitionVersion]] = {
+    def incrementVersion(partitionVersion: PartitionVersion): PartitionVersion =
+      partitionVersion.copy(version = VersionNumber(partitionVersion.version.number + 1))
 
-  override def commit(newVersion: TableVersions.TableUpdate): F[TableVersions.CommitResult] = ???
+    for {
+      tableVersion <- currentVersion(table)
+      nextVersions = partitions.map(
+        partition =>
+          tableVersion.partitionVersions
+            .find(_.partition == partition)
+            .map(incrementVersion)
+            .getOrElse(PartitionVersion(partition, VersionNumber(1))))
+    } yield nextVersions
+  }
+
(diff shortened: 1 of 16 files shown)


core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -26,7 +26,7 @@ trait TableVersions[F[_]] {
     * This performs no checking if data has been written to the associated paths but it will verify that these versions
     * 1) haven't been committed before and 2) these are the next versions to be committed for each of the partitions.
     */
-  def commit(newVersion: TableVersions.TableUpdate): F[CommitResult]
+  def commit(table: TableName, update: TableVersions.TableUpdate): F[CommitResult]
```
stettix, line 29: This was a bit of an omission in the previous version...

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -37,7 +37,7 @@ object TableVersions {
       userId: UserId,
       message: UpdateMessage,
       timestamp: Instant,
-      updatedPartitions: List[PartitionVersion])
+      partitionUpdates: List[PartitionOperation])
```
stettix, line 40: This change was needed to enable partitions to be removed.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -1,16 +1,102 @@
...
+    * Produce current table version based on history of updates.
+    */
+  private[core] def applyUpdate(initial: TableVersion)(operations: List[PartitionOperation]): TableVersion = {
+
+    def applyOp(agg: Map[Partition, VersionNumber], op: PartitionOperation): Map[Partition, VersionNumber] = op match {
+      case AddPartitionVersion(partitionVersion: PartitionVersion) =>
+        agg + (partitionVersion.partition -> partitionVersion.version)
+      case RemovePartition(partition: Partition) =>
+        agg - partition
+    }
```
stettix, line 100: I have a feeling all of this could be simplified using, umm, monoids? 😁 Or some other clever Cats bits. Suggestions welcome!

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -1,16 +1,102 @@
...
+    // Derive current version of a table by folding over the history of changes
+    for {
+      allTableUpdates <- allUpdates.get
+      tableUpdates <- allTableUpdates
+        .get(table)
+        .fold(F.raiseError[List[TableUpdate]](new Exception(s"Table '$table' not found")))(F.pure)
+      operations = tableUpdates.flatMap(_.partitionUpdates)
+    } yield InMemoryTableVersions.applyUpdate(TableVersion.empty)(operations)
+
+  override def nextVersions(table: TableName, partitions: List[Partition]): F[List[PartitionVersion]] = {
```
tomwadeson, line 33: I think this could (and I *think* should) be provided as a default implementation on the trait.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -1,16 +1,102 @@
...
+
+    val applyUpdate: TableUpdates => Either[Exception, TableUpdates] = { currentTableUpdates =>
+      if (currentTableUpdates.contains(table)) {
+        val updated = currentTableUpdates + (table -> (currentTableUpdates(table) :+ update))
+        Right(updated)
+      } else
+        Left(new Exception(s"Unknown table '${table.fullyQualifiedName}'"))
+    }
+
+    allUpdates.modifyEither(applyUpdate).map(_ => SuccessfulCommit)
```
tomwadeson:
```scala
.map(_ => SuccessfulCommit) === .as(SuccessfulCommit)
```

core/src/main/scala/com/gu/tableversions/core/util/RichRef.scala
```diff
@@ -0,0 +1,27 @@
...
+      *
+      * If the provided update function `f` returns a `Right[A]`, the contained value will be used to update the `Ref`.
+      * If `f` returns a `Left[E]`, this will cause the return `F` to be an error.
+      */
+    def modifyEither[E <: Throwable](f: A => Either[E, A]): F[Unit] =
+      underlying
+        .modify(a =>
+          f(a) match {
+            case Left(e)     => (a, e.raiseError[F, Unit])
+            case Right(newA) => (newA, ().pure[F])
```
tomwadeson:
```scala
().pure[F] === F.unit
```

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -53,4 +53,9 @@ object TableVersions {
 
   case class ErrorMessage(value: String) extends AnyVal
 
+  /** ADT for operations on individual partitions. */
+  sealed trait PartitionOperation
```
tomwadeson, line 57: Companion? 😁

tomwadeson (commented):
A few very very tiny comments. Will give more if this is still open in a couple of weeks! Very nice work.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -1,16 +1,102 @@
...
+object InMemoryTableVersions {
+
+  type TableUpdates = Map[TableName, List[TableUpdate]]
+
+  /**
+    * Safe constructor
+    */
+  def apply[F[_]](implicit F: Sync[F]): F[InMemoryTableVersions[F]] =
+    for {
+      ref <- Ref[F].of(Map[TableName, List[TableUpdate]]())
```
tomwadeson:
```scala
Ref[F].of(Map[TableName, List[TableUpdate]]()).map(new InM...)
```
?

core/src/test/scala/com/gu/tableversions/core/TableVersionsSpec.scala
```diff
@@ -0,0 +1,194 @@
...
+trait TableVersionsSpec {
+  this: FlatSpec with Matchers =>
+
+  def tableVersionsBehaviour(emptyTableVersions: IO[TableVersions[IO]]): Unit = {
+
+    val table = TableName("schema", "table")
+    val userId = UserId("Test user")
+    val date = PartitionColumn("date")
+
+    it should "have an idempotent 'init' operation" in {
```
tomwadeson, line 26: Should this assert against the result of the first `.init` *and* subsequent ones?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
tomwadeson: Oh, also, might as well use the `TableUpdates` type alias

core/src/main/scala/com/gu/tableversions/core/util/RichRef.scala
stettix: Ah! You should tell that guy on the cats-effect Gitter... 😆

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Ah yes, I always forget...

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: I'll have a ponder about that. Mind you, with @TBonnin 's suggested changes, which I hope we'll get to try out in the next sprint, the `nextVersions` method will go away altogether,

core/src/test/scala/com/gu/tableversions/core/util/RichRefSpec.scala
```diff
@@ -0,0 +1,36 @@
+package com.gu.tableversions.core.util
+
+import cats.effect.IO
+import cats.effect.concurrent.Ref
+import org.scalatest.{FlatSpec, Matchers}
+
+class RichRefSpec extends FlatSpec with Matchers {
```
tomwadeson, line 7: Think this would be a good candidate for a `TableDriven` thingymajig.

And maybe calling to `.attempt` and asserting on `.isLeft` vs. `.isRight`?

stettix: Perhaps. Can you think of more than the two cases though?

As for `.attempt`, is that to avoid having to catch the exception?

tomwadeson: Nope. `TableDriven` might be overkill then - how about just a couple of assertions in one test body?

`.attempt` was so that you could have a common type in a `Table` for expected results if you went `TableDriven`. But yeah, I'd probably reach for that before `the exception thrownBy` too.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
TBonnin: I am not sure you'll get much more using a monoid. The `combine` part here is already quite simple.

TBonnin (approved):
I concur with @tomwadeson comments but it looks good otherwise.

core/src/test/scala/com/gu/tableversions/core/util/RichRefSpec.scala
stettix: As discussed IRL, I played with this but decided to leave as-is.

PR #7, opened 2019-04-17
Refactor model for snapshot table versions

pvighi:
This PR contains some refactorings in the core model in and around the `TableVersions` trait:

- Introduce `TableVersion` ADT with types for snapshot and partitioned table cases. We found that there are special cases for handling versions for the different types of tables, and encoding this in the types makes things more clear.
- Add `InitTable` operation. We found that this was necessary to let the implementation know which type of table it has in order to return the appropriate `TableVersion` type. But it also seems useful in that the event sequence for a table now fully describes the initialisation of the table as well as version updates.
- Add `AddTableVersion` operation for snapshot tables. This is to distinguish updates to the version of snapshot tables as opposed to updates to versions of partitions in a partitioned table.

### DIFF
--- a/core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
+++ b/core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
@@ -1,11 +1,14 @@
 package com.gu.tableversions.core
 
+import java.sql.Timestamp
+import java.time.Instant
+
 import cats.effect.Sync
 import cats.effect.concurrent.Ref
 import cats.implicits._
 import com.gu.tableversions.core.InMemoryTableVersions.TableUpdates
 import com.gu.tableversions.core.TableVersions.CommitResult.SuccessfulCommit
-import com.gu.tableversions.core.TableVersions.PartitionOperation.{AddPartitionVersion, RemovePartition}
+import com.gu.tableversions.core.TableVersions.PartitionOperation.{AddPartitionVersion, InitTable, RemovePartition}
 import com.gu.tableversions.core.TableVersions._
 import com.gu.tableversions.core.util.RichRef._
 
@@ -15,9 +18,22 @@ import com.gu.tableversions.core.util.RichRef._
 class InMemoryTableVersions[F[_]] private (allUpdates: Ref[F, TableUpdates])(implicit F: Sync[F])
     extends TableVersions[F] {
 
-  override def init(table: TableName): F[Unit] =
+  override def init(
+      table: TableName,
+      isSnapshot: Boolean,
+      userId: UserId,
+      message: UpdateMessage,
+      timestamp: Instant): F[Unit] =
     allUpdates.update { prev =>
-      if (prev.contains(table)) prev else prev + (table -> Nil)
+      if (prev.contains(table)) prev
+      else
+        prev + (table -> List(
+          TableUpdate(
+            userId,
+            message,
+            timestamp = timestamp,
+            partitionUpdates = List(InitTable(table, isSnapshot))
+          )))
     }
 
   override def currentVersion(table: TableName): F[TableVersion] =
@@ -28,7 +44,16 @@ class InMemoryTableVersions[F[_]] private (allUpdates: Ref[F, TableUpdates])(imp
         .get(table)
         .fold(F.raiseError[List[TableUpdate]](new Exception(s"Table '$table' not found")))(F.pure)
       operations = tableUpdates.flatMap(_.partitionUpdates)
-    } yield InMemoryTableVersions.applyUpdate(TableVersion.empty)(operations)
+    } yield
+      if (isSnapShot(operations))
+        InMemoryTableVersions.latestSnapshotTableVersion(operations)
+      else
+        InMemoryTableVersions.applyPartitionUpdates(PartitionedTableVersion(Map.empty))(operations)
+
+  def isSnapShot(operations: List[PartitionOperation]) = operations match {
+    case InitTable(_, isSnapshot) :: _ => isSnapshot
+    case _                             => throw new IllegalArgumentException("First operation should be InitTable")
+  }
 
   override def nextVersions(table: TableName, partitions: List[Partition]): F[Map[Partition, Version]] = {
     def maxVersions(operations: List[(Partition, Version)]): Map[Partition, Version] =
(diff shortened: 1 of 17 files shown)


core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -27,39 +42,62 @@ class InMemoryTableVersions[F[_]] private (allUpdates: Ref[F, TableUpdates])(imp
...
+        InMemoryTableVersions.latestSnapshotTableVersion(operations)
+      else
+        InMemoryTableVersions.applyPartitionUpdates(PartitionedTableVersion(Map.empty))(operations)
+
+  def isSnapShot(operations: List[TableOperation]) = operations match {
+    case InitTable(_, isSnapshot) :: _ => isSnapshot
+    case _                             => throw new IllegalArgumentException("First operation should be InitTable")
   }
 
+  override def nextVersions(table: TableName, partitions: List[Partition]): F[Map[Partition, Version]] =
```
stettix, line 57: This method has become even uglier than before as part of these changes, but we expect this method to disappear when we do the "UUID paths" change so we're not too worried about it...

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -53,12 +53,14 @@ object TableVersions {
...
+  sealed trait TableOperation
 
-  object PartitionOperation {
-    final case class AddPartitionVersion(partition: Partition, version: Version) extends PartitionOperation
-    final case class RemovePartition(partition: Partition) extends PartitionOperation
+  object TableOperation {
+    final case class InitTable(tableName: TableName, isSnapshot: Boolean) extends TableOperation
+    final case class AddTableVersion(version: Version) extends TableOperation
+    final case class AddPartitionVersion(partition: Partition, version: Version) extends TableOperation
+    final case class RemovePartition(partition: Partition) extends TableOperation
```
stettix, line 63: This is one of the key changes.

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -68,12 +68,6 @@ final case class TableDefinition(name: TableName, location: URI, partitionSchema
...
-object TableVersion {
-
-  def snapshotVersion(version: Version): TableVersion = TableVersion(Map(Partition.snapshotPartition -> version))
-
-  val empty = TableVersion(Map.empty)
-
-}
+sealed trait TableVersion
+final case class PartitionedTableVersion(partitionVersions: Map[Partition, Version]) extends TableVersion
+final case class SnapshotTableVersion(version: Version) extends TableVersion
```
stettix, line 73: This is the other key change.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -15,9 +17,22 @@ import com.gu.tableversions.core.util.RichRef._
 class InMemoryTableVersions[F[_]] private (allUpdates: Ref[F, TableUpdates])(implicit F: Sync[F])
```
tomwadeson, line 17: Disclaimer: sorry, I know it's not something that's been introduced in this PR, but...

We've got quite a bit of business logic in this concrete implementation that looks likely to me to be the only sensible implementation of this API. I think we can hoist it up to `TableVersions` and achieve a few benefits for doing so:

```scala
/**
  * This defines the interface for querying and updating table version information tracked by the system.
  */
trait TableVersions[F[_]] {

  /**
    * Start tracking version information for given table.
    * This must be called before any other operations can be performed on this table.
    */
  def init(table: TableName, isSnapshot: Boolean, userId: UserId, message: UpdateMessage, timestamp: Instant): F[Unit]

  /** Get a description of which version to write to next for the given partitions of a table. */
  def nextVersions(table: TableName, partitions: List[Partition]): F[Map[Partition, Version]]

  /**
    * Update partition versions to the given versions.
    * This performs no checking if data has been written to the associated paths but it will verify that these versions
    * 1) haven't been committed before and 2) these are the next versions to be committed for each of the partitions.
    */
  def commit(table: TableName, update: TableVersions.TableUpdate): F[CommitResult]

  def updates: fs2.Stream[F, TableUpdate]

  // Implementors are free to provide a more efficient override
  def updates(forTable: TableName): fs2.Stream[F, TableUpdate] =
    updates.filter(_.table == forTable)

  /** Get details about partition versions in a table. */
  def currentVersion(table: TableName): F[TableVersion] =
    updates(table).fold(doMyThingWithTableOperations)
}
```

Now implementors are only required to provide the primitives (`updates` and potentially `updates(forTable)`), and you're able to give them a useful bit of the API for free (`currentVersion`).

Plus it doesn't seem unreasonable to me to expose the updates; I'm sure it's useful. And by using an fs2 `Stream`, there's no requirement now to pass potentially very large lists of updates around strictly.

See what you think - I'm just thinking out loud here and it certainly wouldn't need to go in with this PR regardless!

stettix: I like the idea of using fs2 `Stream`s instead of `List`s for these. I'll add that to the list!

W.r.t. pulling out potential common code, I like the suggestion, though I'd probably wait until we did add another implementation.

Regarding your specific suggested code - I'm not sure it makes sense to make implementors implement a stream of all update events across tables? This wouldn't really be used for anything as it stands (other than the non-efficient default implementation of `updates(forTable: TableName)`, and it may be painful to do for some backends. For example if we have an implementation using DynamoDB with a separate document per table as its schema, it would have to do a query per table for to provide this.

tomwadeson: > W.r.t. pulling out potential common code, I like the suggestion, though I'd probably wait until we did add another implementation.

That's cool, no rush!

FTR, that's not really my motivation here though (the pulling out of potential common code). 

I think it would make `TableVersions` more useful and cohesive as an abstraction: the operations defined on it would have stronger relationships with one another and it would provide better guarantees.

> I'm not sure it makes sense to make implementors implement a stream of all update events across tables?

That's fair. Maybe just `updates(forTable: TableName): Stream[F, TableUpdate]` then.

tomwadeson (approved):
Nice work.

PR #8, opened 2019-04-18
Use UUID strings for versions instead of incremental numbers

stettix:
This changes `Version`s to contain opaque strings instead of incremental numbers. We define a format for version strings that contains a timestamp and a UUID. We also remove the `nextVersion` method on the `TableVersion` interface as clients can now generate their own `Version`s as needed.

This means that written partitions will be completely immutable as code will generate new `Version`s every time it attempts to write data. It also ensures two concurrent clients can't attempt to write to the same path.


TBonnin (approved):
looks very good 👍

PR #9, opened 2019-04-18
Remove usage of empty Partition value to handle snapshot table case

stettix:
We used to indicate a snapshot table by having a `Partition` with an empty list of columns. It turns out that this wasn't really useful as we mostly needed special cases to handle snapshot vs. partitioned tables anyway. So we've removed this 'magic' value and have separate functions for handling different cases instead.

### DIFF
--- a/core/src/main/scala/com/gu/tableversions/core/model.scala
+++ b/core/src/main/scala/com/gu/tableversions/core/model.scala
@@ -12,16 +12,15 @@ import cats.effect.IO
   */
 final case class Partition(columnValues: List[Partition.ColumnValue]) {
 
+  require(columnValues.nonEmpty, "Must have at least one partition column")
+
   /** Given a base path for the table, return the path to the partition. */
   def resolvePath(tableLocation: URI): URI = {
     def normalised(dir: URI): URI = if (dir.toString.endsWith("/")) dir else new URI(dir.toString + "/")
-    if (this == Partition.snapshotPartition) {
-      tableLocation
-    } else {
-      val partitionsSuffix =
-        columnValues.map(columnValue => s"${columnValue.column.name}=${columnValue.value}").mkString("", "/", "/")
-      normalised(tableLocation).resolve(partitionsSuffix)
-    }
+
+    val partitionsSuffix =
+      columnValues.map(columnValue => s"${columnValue.column.name}=${columnValue.value}").mkString("", "/", "/")
+    normalised(tableLocation).resolve(partitionsSuffix)
   }
 }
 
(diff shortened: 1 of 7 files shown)


gr211 (approved):

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -12,16 +12,15 @@ import cats.effect.IO
   */
 final case class Partition(columnValues: List[Partition.ColumnValue]) {
 
+  require(columnValues.nonEmpty, "Must have at least one partition column")
+
```
TBonnin, line 15: Can we use `NonEmptyList` rather than relying on runtime check?

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -77,42 +86,31 @@ object VersionedDataset {
...
+    val partitionColumnsList = partitionSchema.columns.map(_.name)
+    val partitionsDf = dataset.selectExpr(partitionColumnsList: _*).distinct()
+    val partitionRows = partitionsDf.collect().toList
+
+    def rowToPartition(row: Row): Partition = {
+      val partitionColumnValues: List[(Partition.PartitionColumn, String)] =
+        partitionSchema.columns zip row.toSeq.map(_.toString)
+
+      val columnValues: List[Partition.ColumnValue] = partitionColumnValues.map {
+        case (partitionColumn, value) => Partition.ColumnValue(partitionColumn, value)
```
gr211: you could rewrite to
```
val columnValues: List[Partition.ColumnValue] = 
   partitionColumnValues map Partition.ColumnValue.tupled
``

File: core/src/main/scala/com/gu/tableversions/core/model.scala (hunk 1 above)
Comment (stettix) [reply core/src/main/scala/com/gu/tableversions/core/model.scala:15]: Yes we talked about that, we think there are a number of places we can do this but we decided it would be a step to far for this refactoring.

Comment (stettix) [reply core/src/main/scala/com/gu/tableversions/core/model.scala:15]: @TBonnin On reflection, it seemed better to include this refactoring here so we did as you suggested.

File: spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala [hunk 3]
```diff
@@ -77,42 +86,32 @@ object VersionedDataset {
...
+      val partitionColumnValues: List[(Partition.PartitionColumn, String)] =
+        partitionSchema.columns zip row.toSeq.map(_.toString)
 
-      partitionRows.map(rowToPartition)
+      val columnValues = partitionColumnValues.map(Partition.ColumnValue.tupled)
+
+      columnValues match {
+        case head :: tail => Partition(head, tail: _*)
+        case _            => throw new Exception("Empty list of partitions not valid for partitioned table")
+      }
```
stettix, line 111: This runtime error is needed because `PartitionSchema` is used for both the partitioned and snapshot cases at the moment hence the type system doesn't enforce an empty list of partition columns. One for a later refactoring, we think...

PR #10, opened 2019-04-23
Refactor Version type to require a defined timestamp + UUID format

stettix:
This is a refactoring to specify more strictly the definition of a `Version`. Instead of just a `String`, it can now only be created with a timestamp and UUID. Also, the parsing and rendering of string representations of versions is moved to the `Version` type.

### DIFF
--- a/core/src/main/scala/com/gu/tableversions/core/Version.scala
+++ b/core/src/main/scala/com/gu/tableversions/core/Version.scala
@@ -1,0 +1,60 @@
+package com.gu.tableversions.core
+
+import java.time.{Instant, LocalDateTime, ZoneId}
+import java.time.format.DateTimeFormatter
+import java.util.UUID
+
+import cats.effect.IO
+import cats.syntax.either._
+
+import scala.util.Try
+
+/**
+  * Type that represents a valid version, as used by tables and partitions.
+  */
+final case class Version(timestamp: Instant, uuid: UUID) {
+
+  import Version._
+
+  def label: String = {
+    val versionString = uuid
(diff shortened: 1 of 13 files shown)


core/src/main/scala/com/gu/tableversions/core/Version.scala
```diff
@@ -0,0 +1,60 @@
...
+object Version {
+
+  val Unversioned = Version(Instant.MIN, new UUID(0L, 0L))
+
+  /** Generator for versions using a timestamp + UUID format */
+  implicit val generateVersion: IO[Version] =
+    IO { Version(Instant.now(), UUID.randomUUID()) }
+
+  /** Regex that can be used to recognise a valid format string. */
+  val TimestampAndUuidRegex = """(\d{8}-\d{6}.\d{9})-(.*)""".r
```
TBonnin, line 37: we have added nanoseconds 👍

TBonnin (approved):
lgtm

core/src/main/scala/com/gu/tableversions/core/Version.scala
stettix: Yes... for non-obvious reasons...

We only had seconds at first, but then we realised we need to reconstruct a `Version` object from the string label. So the choice was either including the nanoseconds in the string label, or make sure that the `Version` object didn't have nanosecond values in their timestamps - otherwise the roundtrip of object -> string label -> object didn't result in the original object!

So we did the former, still not sure if that's ideal but...

PR #11, opened 2019-04-25
Add rollback support

stettix:
This PR adds the ability to list the version history for a table, and rollback (and roll forward) functionality as in the ability to select a version from the history to be the current version. This functionality is exercised via the end-to-end specs in the `example` module.

In addition, this also contains some refactoring changes to the version handling code, including:

* Pulling common code from the concrete `InMemoryTableVersions` class into the `TableVersions` trait so that the concrete implementation only has to deal with aspects specific to its storage and retrieval of version data. This was suggested by @tomwadeson [in a previous PR](https://github.com/guardian/table-versions/pull/7#discussion_r276214981) and helps separate business logic from the storage backend concerns, and providing stronger guarantees from the `TableVersion` abstraction.
* Using a single `TableLoader` class in the `example` module, removing the duplication between the three different classes that existed before.
* Removing the `CommitResult` type as after the earlier UUID path changes, the `commit` operation performs no validation on its parameters.

Due to the refactoring changes, it might be easier to review the changes commit by commit!


build.sbt
```diff
@@ -66,4 +67,5 @@ lazy val examples = project
     libraryDependencies ++= sparkDependencies
   )
   .settings(parallelExecution in Test := false)
+  .settings(fork in Test := true)
```
stettix, line 70: I added these to the 'spark' and 'examples' modules in the hope that this would reduce the number of times I have to restart `sbt` due to out of memory errors while developing.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -13,39 +17,93 @@ trait TableVersions[F[_]] {
...
+      operations = updatesForCurrentVersion.flatMap(_.operations)
+    } yield
+      if (isSnapshotTable(operations))
+        latestSnapshotTableVersion(operations)
+      else
+        applyPartitionUpdates(PartitionedTableVersion(Map.empty))(operations)
+
+  /** Return the history of table updates, most recent first. */
+  def updates(table: TableName)(implicit F: Sync[F]): F[List[TableUpdateHeader]] =
+    tableState(table).map(_.updates.map(_.header).reverse)
```
stettix: This is one of the new operations, getting the version history of a table.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -13,39 +17,93 @@ trait TableVersions[F[_]] {
...
-    * This performs no checking if data has been written to the associated paths but it will verify that these versions
-    * 1) haven't been committed before and 2) these are the next versions to be committed for each of the partitions.
     */
-  def commit(table: TableName, update: TableVersions.TableUpdate): F[CommitResult]
+  def commit(table: TableName, update: TableVersions.TableUpdate): F[Unit]
+
+  /**
+    * Set the current version of a table to refer to an existing version.
+    */
+  def setCurrentVersion(table: TableName, id: TableVersions.CommitId): F[Unit]
```
stettix, line 55: This is the other new operation: selecting a previous version as the current version.

core/src/test/scala/com/gu/tableversions/core/TableVersionsSpec.scala
```diff
@@ -187,38 +188,249 @@ trait TableVersionsSpec {
...
       initialTableVersion shouldBe SnapshotTableVersion(Version.Unversioned)
-
-      commitResult1 shouldBe SuccessfulCommit
       currentVersion1 shouldBe tableVersion1
-
-      commitResult2 shouldBe SuccessfulCommit
       currentVersion2 shouldBe currentVersion2
     }
 
+    it should "allow table versions of a partitioned table to be updated from version history" in {
```
stettix, line 198: These are the unit tests for the new functionality (roll back and forward) in `TableVersions`.

spark/src/test/scala/com/gu/tableversions/spark/VersionedDatasetSpec.scala
```diff
@@ -205,29 +198,6 @@ class VersionedDatasetSpec extends FlatSpec with Matchers with SparkHiveSuite {
...
-        timestamp: Instant): IO[Unit] =
-      IO.unit
-
-    override def currentVersion(table: TableName): IO[TableVersion] =
-      IO(currentVersions(table))
-
-    override def commit(table: TableName, newVersion: TableVersions.TableUpdate): IO[TableVersions.CommitResult] =
-      committedTableUpdatesRef.update(_ :+ table -> newVersion).map(_ => CommitResult.SuccessfulCommit)
-  }
-
```
stettix, line 230: After refactoring, the `InMemoryTableVersions` class is so simple that the stub isn't much simpler - hence we're ditching this stub and just using the `InMemoryTableVersions` implementation instead in these tests.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -13,39 +17,93 @@ trait TableVersions[F[_]] {
...
 
-  /** Result type for commit operation */
-  sealed trait CommitResult
+  final case class UpdateMessage(content: String) extends AnyVal
 
-  object CommitResult {
-    case object SuccessfulCommit extends CommitResult
-    final case class InvalidCommit(invalidPartitions: Map[Partition, ErrorMessage]) extends CommitResult
+  /** A collection of updates to partitions to be applied and tracked as a single atomic change. */
+  final case class TableUpdate(header: TableUpdateHeader, operations: List[TableOperation])
```
TBonnin: Why introducing the concept of TableUpdate? Isn't a `List[TableOperation]` not enough? What does it bring?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -17,95 +13,63 @@ import com.gu.tableversions.core.util.RichRef._
...
-        prev + (table -> List(
-          TableUpdate(
-            userId,
-            message,
-            timestamp = timestamp,
-            operations = List(InitTable(table, isSnapshot))
-          )))
+  override def commit(table: TableName, update: TableVersions.TableUpdate): F[Unit] = {
+    val applyUpdate: TableUpdates => Either[Exception, TableUpdates] = { currentTableUpdates =>
+      if (currentTableUpdates.contains(table)) {
```
TBonnin: Can you use `Map.get().fold()()` here rather than `if`
ie:
```
currentTableUpdates
.get(table)
.fold {
    Left(unknownTableError(table)
} { oldState =>
    val newState = TableState(
        currentVersion = update.header.id,
        updates = oldState.updates :+ update
    )
    Right(currentTableUpdates + (table -> newState))
}
```

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -17,95 +13,63 @@ import com.gu.tableversions.core.util.RichRef._
...
+        case Some(currentTableState) =>
+          if (currentTableState.updates.exists(_.header.id == id)) {
+            val newTableState = currentTableState.copy(currentVersion = id)
+            val updated = currentTableUpdates + (table -> newTableState)
+            Right(updated)
+          } else
+            Left(unknownCommitId(id))
+
+        case None => Left(unknownTableError(table))
+      }
```
TBonnin, line 40: Pattern matching instead of `fold`. Any specific reason?

build.sbt
gr211: why not add it to and use the `testSettings` val ?

stettix: I don't want it on the non-Spark projects as it's slower and those projects don't have the issue.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -17,95 +13,63 @@ import com.gu.tableversions.core.util.RichRef._
...
-      else
-        prev + (table -> List(
-          TableUpdate(
-            userId,
-            message,
-            timestamp = timestamp,
-            operations = List(InitTable(table, isSnapshot))
-          )))
+  override def commit(table: TableName, update: TableVersions.TableUpdate): F[Unit] = {
+    val applyUpdate: TableUpdates => Either[Exception, TableUpdates] = { currentTableUpdates =>
```
gr211, line 17: how about match case style rather than if/else ?
```
case currentTableUpdates if currentTableUpdates contains table =>
        val updated = currentTableUpdates + (table -> TableState(
          currentVersion = update.header.id,
          updates = currentTableUpdates(table).updates :+ update))
        Right(updated)
case _ =>
        Left(unknownTableError(table))
```

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -60,4 +118,47 @@ object TableVersions {
     final case class RemovePartition(partition: Partition) extends TableOperation
   }
 
+  /** The current state of the table, including all updates and the current version. */
+  case class TableState(currentVersion: CommitId, updates: List[TableUpdate])
```
gr211: Would it be safer/clearer to use a data structure a bit more contraining than `List`, like a `SortedSet` or something to guarantee updates order ? The companion object to `TableSate` could still take a `List` for construction yet `updates` would be a sorted thing.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -17,95 +13,63 @@ import com.gu.tableversions.core.util.RichRef._
...
-    } yield
-      if (isSnapShot(operations))
-        InMemoryTableVersions.latestSnapshotTableVersion(operations)
-      else
-        InMemoryTableVersions.applyPartitionUpdates(PartitionedTableVersion(Map.empty))(operations)
-
-  private def isSnapShot(operations: List[TableOperation]) = operations match {
-    case InitTable(_, isSnapshot) :: _ => isSnapshot
-    case _                             => throw new IllegalArgumentException("First operation should be InitTable")
+    allUpdates.modifyEither(applyUpdate).void
```
gr211: naive question. What purpose does `.void` have seeing `modifyEither(...)` return type is `F[Unit]` also?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -17,95 +13,63 @@ import com.gu.tableversions.core.util.RichRef._
 class InMemoryTableVersions[F[_]] private (allUpdates: Ref[F, TableUpdates])(implicit F: Sync[F])
```
gr211, line 13: General question: is it an anticipated concern that within the same jvm process 2 or more concurrent operations might modify `TableUpdates` ? Or more a general design direction ?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: Not naive - that is indeed unneccessary! 😊

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: No particular reason - I'm just not in the habit of using `fold` on `Option`s. But I can change...

stettix: BTW do you also prefer `fold` to `map` + `getOrElse`?

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Not sure I understand - `List` guarantees the order?

stettix: Oh you mean, use a `Set` to avoid duplicates but use `SortedSet` to still retain order?

Maybe - I have a feeling that this `List` may become an `fs2.Stream` at some point though, as you often don't care about the full history of the table, just the most recent changes. (See @tomwadeson 's comment linked elsewhere). 

But on that note - the `TableState` type isn't part of the public interface for the version repository, so I should make it protected. At which point, presumably the question of the exact type of this is slightly less important?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: I don't know. Hence erring on the side of caution...

stettix: Also note that this impl will only be used in tests...

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
TBonnin: I don't mind to be honest, as long as it is consistant throughout the codebase

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
```diff
@@ -17,95 +13,63 @@ import com.gu.tableversions.core.util.RichRef._
...
+    allUpdates.modifyEither(applyUpdate).void
   }
 
+  override def tableState(table: TableName): F[TableState] =
+    for {
+      allTableUpdates <- allUpdates.get
+      tableState <- allTableUpdates
+        .get(table)
+        .fold(F.raiseError[TableState](unknownTableError(table)))(F.pure)
+    } yield tableState
```
gr211, line 50: Not a request for change. Just a stylistic question on the back of that other discussion about `Ref` and exceptions.

Seeing we're in `F` here (unlike yesterday we were still in `Ref`) and the same law `F.delay[A](throw e) <-> F.raiseError(e)` should apply for `IO`, is the following easier to read, as in 
* no `Option.fold()` 
* only 1 explicit call to `F.(...)` in lieu of 2, 
*  1 line shorter

```
for {
      allTableUpdates <- allUpdates.get
      tableState <- F.delay(allTableUpdates(table)).adaptError {
        case _ => unknownTableError(table)
      }
    } yield tableState
```

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
gr211: Just saw that, re. the point about `updates` being an explicitely sorted structure of sort

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
gr211: The suffix `Header` induced me to wrongly thinking it had something to do with a table's header, or maybe schema of sort.
How about `TableOperationMetadata`, so it semantically refers to the name of the `operations` type ?

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: Actually, since I wrote that I have learnt that there is an `F.fromOption()` method which would simplify this. In which case this becomes:

```scala
    for {
      allTableUpdates <- allUpdates.get
      tableState <- F.fromOption(allTableUpdates.get(table), unknownTableError(table))
    } yield tableState
```

Which is nice and simple?

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
gr211: or as discussed IRL, a comment saying it is to the discretion of concrete implementations to sort the updates most recent first or smtg

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: True, I just found that 'metadata' is quite heavily used, e.g. `TableMetadata` (albeit not in this repo) along with "metastore".

Not sure about `TableOperationMetadata` as it's metadata pertaining to the table update, not the operations... how about `TableUpdateMetadata`?

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: Yes I think along with adding an explicit test for this in `TableVersionsSpec` would be useful.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
gr211: I think `getOrElse` would usually attract the ire of fp fanatics ;-)

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
gr211: That's beautiful ;-)

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: Why? `Option.get` is obviously bad, but what's the problem with `getOrElse`?

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
gr211: Works for me +1

gr211 (approved):

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
TBonnin: Unlike `fold` it doesn't force the return type to be the same as the `Option`.
```
scala> Some(1).getOrElse("something")
res0: Any = 1

scala> Some(1).fold("something")(identity)
<console>:12: error: type mismatch;
 found   : Int
 required: String
       Some(1).fold("something")(identity)
```

stettix: That's all the justification I need, `fold` it is then!

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
gr211: Could this be a case of defensive coding ? No need to change change as it's for test only. Just that _erring onthe side of caution_ makes it sound a bit like that.

stettix: Perhaps, that is a fair point. But I think it's reasonable to implement everything that involves mutable state in such a way that it works correctly however it's used. On balance, I'd rather not spend the time trying to reason about whether it's safe or not, making assumptions about how it will be used in the future.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: Can't tell if sarcasm or not... 😁

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
stettix: The other field, `header`? Which includes the commit ID, as well as details about who performed the update, at what time. I.e. metadata about the update (as it will soon be renamed to).

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: Just to mention, it doesn't infer the right type with the above code, so you have to explicitly specify the type for the `fold`, which is a minor annoyance. I.e. in this case you have to do `.fold[Either[Exception, TableUpdates]`. Sad!

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
stettix: As mentioned above, this doesn't work because it doesn't infer the right type:

```scala
Some(1).fold(Left("Empty!"))(n => Right(n + 2))
```

So you have to explicitly provide the type:

```scala
Some(1).fold[Either[String, Int]](Left("Empty!"))(n => Right(n + 2))
```

Unless I'm missing something? (I'm still going with the `fold` anyway)

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
gr211: It wasn't. `F.fromOption` is the better version here

core/src/test/scala/com/gu/tableversions/core/TableVersionsSpec.scala
```diff
@@ -187,38 +191,278 @@ trait TableVersionsSpec {
...
+      versionAfterWrites shouldBe SnapshotTableVersion(version3)
+      versionAfterWrites2 shouldEqual versionAfterWrites
+
+      versionSetToInitial shouldBe SnapshotTableVersion(Version.Unversioned)
+      versionSetToFirst shouldBe SnapshotTableVersion(version1)
+      versionSetToSecond shouldBe SnapshotTableVersion(version2)
+      versionSetToLatest shouldEqual versionAfterWrites
+    }
+
+    it should "return updates in the same order as they were committed, but with the most recent first" in {
```
stettix, line 379: Now with added tests for the ordering of the update history returned by the implementation.

core/src/main/scala/com/gu/tableversions/core/TableVersions.scala
```diff
@@ -13,42 +17,96 @@ trait TableVersions[F[_]] {
...
+  /**
+    * Get details about partition versions in a table.
+    */
+  def currentVersion(table: TableName)(implicit F: Sync[F]): F[TableVersion] =
+    // Derive current version of a table by folding over the history of changes
+    // until either the latest or version marked as 'current' is reached.
+    for {
+      ts <- tableState(table)
+      matchingUpdates = ts.updates.span(_.metadata.id != ts.currentVersion)
+      updatesForCurrentVersion = matchingUpdates._1 ++ matchingUpdates._2.take(1)
```
pvighi, line 35: would replacing matchingUpdates with separate descriptive names make this clearer?
something like this (but with better names probably) :
```scala
      (previousUpdates, currentUpdates) = ts.updates.span(_.metadata.id != ts.currentVersion)
      updatesForCurrentVersion = previousUpdates ++ currentUpdates.take(1)
````

Comment (stettix) [reply core/src/main/scala/com/gu/tableversions/core/TableVersions.scala:35]: Yes indeed! I thought this wasn't possible due to the "missing `withFilter`" issue but I guess this is OK because it's inside a `map` not a `flatMap`. Well spotted!

File: core/src/main/scala/com/gu/tableversions/core/TableVersions.scala [hunk 16]
```diff
@@ -13,42 +17,96 @@ trait TableVersions[F[_]] {
...
-  final case class UpdateMessage(content: String) extends AnyVal
+  object TableUpdateMetadata {
+
+    def apply(userId: UserId, message: UpdateMessage, timestamp: Instant): TableUpdateMetadata =
+      TableUpdateMetadata(createId(), userId, message, timestamp)
+
+    private def createId(): CommitId = CommitId(UUID.randomUUID().toString)
+  }
+
+  final case class CommitId(id: String) extends AnyVal
```
pvighi, line 90: maybe commitId could have a UUID inside instead of a string ?
for consistency with the version class

stettix: I had a think about this and I'm tempted to leave it as it is. It's less important here, as we're not using the string version of the ID to write file paths etc that will be parsed back later.

core/src/main/scala/com/gu/tableversions/core/InMemoryTableVersions.scala
TBonnin: I don't think you are missing anything. Type inference in Scala is quite limited

PR #12, opened 2019-04-26
Add AWS Glue metastore support

pvighi:
We currently only have a Hive Metastore implementation this PR adds an AWS Glue implementation.
Unfortunately, to test this properly we need to run it against a real Glue instance so only integration tests are provided.

Side note: It looks like the IntelliJ test runner doesn't support this test configuration so the integration tests can only  be run by doing it:test in the sbt shell


metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
```diff
@@ -35,4 +37,35 @@ object VersionPaths {
       new URI(versioned(normalised(partitionPath.toString)))
     }
 
+  /**
```
pvighi, line 40: These two functions are moved here from the SparkHiveMetastore but are useful for both metastore implementations. I though they fit in here

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+  val glue: AWSGlue = AWSGlueClient.builder().withCredentials(credentials).withRegion(region).build()
+  val schema = "temp"
+  def tableLocation(tableName: String) = new URI(s"s3://ophan-temp-schema/test-data/$tableName")
+
+  val snapshotTable =
+    TableDefinition(TableName(schema, "test_snapshot"), tableLocation("test_snapshot"), PartitionSchema.snapshot)
+
+  val partitionedTable = TableDefinition(TableName(schema, "test_partitioned"),
+                                         tableLocation("test_partitioned"),
+                                         PartitionSchema(List(PartitionColumn("date"))))
```
pvighi: I'm just thinking that maybe I could add a random component to the table names so that multiple people running this test concurrently don't interfere with each other.

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+import java.net.URI
+
+import cats.effect.IO
+import com.amazonaws.auth.profile.ProfileCredentialsProvider
+import com.amazonaws.auth.{
+  AWSCredentialsProviderChain,
+  EnvironmentVariableCredentialsProvider,
+  InstanceProfileCredentialsProvider,
+  SystemPropertiesCredentialsProvider
+}
```
stettix: `import com.amazonaws.auth._` perhaps..?

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+import cats.implicits._
+
+class GlueMetastoreSpec extends FlatSpec with Matchers with BeforeAndAfterAll with MetastoreSpec {
+
+  lazy val region = Regions.EU_WEST_1
+
+  lazy val credentials = new AWSCredentialsProviderChain(
+    new EnvironmentVariableCredentialsProvider(),
+    new SystemPropertiesCredentialsProvider(),
+    new ProfileCredentialsProvider("ophan"),
```
stettix: I'm not sure about this, as it is Guardian specific... is your current plan to run this test from TeamCity?

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
stettix: Good idea!

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,184 @@
...
+  GetPartitionsResult,
+  GetTableRequest,
+  PartitionInput,
+  StorageDescriptor,
+  TableInput,
+  UpdatePartitionRequest,
+  UpdateTableRequest,
+  Partition => GluePartition,
+  Table => GlueTable
+}
```
stettix: Is it not possible to do an `import com.amazonaws.services.glue.model._` as you're aliasing some of the imports?

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
```diff
@@ -180,8 +182,8 @@ trait MetastoreSpec {
 
       } yield ()
 
-      val ex = the[Exception] thrownBy scenario.unsafeRunSync()
-      ex.getMessage.toLowerCase should include regex "partition not found"
+      val ex = the[Exception] thrownBy scenario.guarantee(tearDownUnderlyingTable).unsafeRunSync()
+      ex.getMessage.toLowerCase should include regex "notfound|not found"
```
stettix: Any particular reason we want the error messages to be different for the two implementations?

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
```diff
@@ -57,15 +58,16 @@ trait MetastoreSpec {
 
   def metastoreWithPartitionsSupport(
       emptyMetastore: IO[Metastore[IO]],
-      initHiveTable: IO[Unit],
-      table: TableName): Unit = {
+      initUnderlyingTable: IO[Unit],
+      table: TableName,
+      tearDownUnderlyingTable: IO[Unit] = IO { () }): Unit = {
```
stettix: There is an `IO.unit` constant that you can use instead of `IO { () }`.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
```diff
@@ -45,7 +46,7 @@ trait MetastoreSpec {
       } yield (initialVersion, version1, firstUpdatedVersion, version2, secondUpdatedVersion, revertedVersion)
 
       val (initialVersion, version1, firstUpdatedVersion, version2, secondUpdatedVersion, revertedVersion) =
-        scenario.unsafeRunSync()
+        scenario.guarantee(tearDownUnderlyingTable).unsafeRunSync()
```
stettix, line 49: TIL about `guarantee` - nice! 👍

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
```diff
@@ -17,13 +17,14 @@ trait MetastoreSpec {
 
   def metastoreWithSnapshotSupport(
       emptyMetastore: IO[Metastore[IO]],
-      initHiveTable: IO[Unit],
-      table: TableName): Unit = {
+      initUnderlyingTable: IO[Unit],
+      table: TableName,
+      tearDownUnderlyingTable: IO[Unit] = IO { () }): Unit = {
```
stettix: Nice renaming! 👍

stettix: You can use `IO.unit` here as well btw.

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,77 @@
...
+    val storageDescription = new StorageDescriptor()
+      .withLocation(table.location.toString)
+      .withColumns(column("id", "String"), column("field1", "String"))
+
+    val input = {
+      val unpartitionedInput = new TableInput()
+        .withName(table.name.name)
+        .withDescription("table used in integration tests for table versions")
+        .withStorageDescriptor(storageDescription)
+      if (table.isSnapshot) unpartitionedInput else unpartitionedInput.withPartitionKeys(column("date", "date"))
```
stettix: You probably want to have different names for the two partition columns.

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,184 @@
...
+import com.gu.tableversions.core.Partition.{ColumnValue, PartitionColumn}
+import com.gu.tableversions.core.{PartitionedTableVersion, SnapshotTableVersion, TableName, TableVersion, _}
+import com.gu.tableversions.metastore.Metastore.TableOperation
+import com.gu.tableversions.metastore.Metastore.TableOperation.{
+  AddPartition,
+  RemovePartition,
+  UpdatePartitionVersion,
+  UpdateTableVersion
+}
```
stettix: `import ... ._` ... (sorry!)

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,184 @@
...
+
+    for {
+      glueTable <- getGlueTable(table)
+      partitionColumns <- F.delay(getPartitionColumns(glueTable))
+      tableLocation <- F.delay(findTableLocation(glueTable))
+      version <- if (partitionColumns.isEmpty)
+        F.delay(snapshotTableVersion(tableLocation))
+      else
+        getPartitionedTableVersion(table, tableLocation, partitionColumns)
+    } yield (version)
```
stettix: Unnecessary parantheses around `version`...? #pedantic

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,184 @@
...
+  override def update(table: TableName, changes: Metastore.TableChanges): F[Unit] =
+    changes.operations.traverse_(appliedOp(table))
+
+  private def appliedOp(table: TableName)(operation: TableOperation): F[Unit] =
+    operation match {
+      case AddPartition(partition, version)           => addPartition(table, partition, version)
+      case UpdatePartitionVersion(partition, version) => updatePartitionVersion(table, partition, version)
+      case RemovePartition(partition)                 => removePartition(table, partition)
+      case UpdateTableVersion(versionNumber)          => updateTableLocation(table, versionNumber)
+    }
```
stettix, line 74: It's interesting that the two methods above are the same for this implementation and the Spark/Hive one. We could consider pulling this out into the base trait in a later PR, similar to what I did for `TableVersions`.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
pvighi: that is nicer, I'll change it

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,184 @@
...
+      storageDescriptor <- F.delay(new StorageDescriptor().withLocation(location))
+      partitionValues <- F.delay(partition.columnValues.map(_.value).toList)
+      input <- F.delay(new PartitionInput().withValues(partitionValues).withStorageDescriptor(storageDescriptor))
+
+      addPartitionRequest = new CreatePartitionRequest()
+        .withDatabaseName(table.schema)
+        .withTableName(table.name)
+        .withPartitionInput(input)
+
+      _ <- F.delay { glue.createPartition(addPartitionRequest) }
```
stettix, line 95: Maybe worth pulling out an `addPartition` method to match the other cases, e.g. `removePartition` etc? Whatever you prefer anyway.

stettix: (Actually, I misread the rest of the code, so just ignore the above comment!)

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,184 @@
...
+
+    def partitionLocation(tableLocation: URI): String = {
+      val unversionedLocation: String = partition.resolvePath(tableLocation).toString
+      if (version == Version.Unversioned) unversionedLocation
+      else
+        unversionedLocation + version.label
+    }
+    for {
+      glueTable <- getGlueTable(table)
+      tableLocation <- F.delay(findTableLocation(glueTable))
```
stettix: If `findTableLocation` is a pure function, you could do this instead here:

```scala
tableLocation = F.pure(findTableLocation(glueTable))
```

Or, if it's not pure (i.e. if it can throw exceptions) maybe it's worth moving the `F.delay` to the implementation of `findTableLocation`?

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
pvighi: I just didn't bother to catch the aws exception and throw a custom one.
Since this didn't check the specific error i assumed we didn't care as long as it said not found somewhere. The aws exception says entitynotfound or something like that without spaces
But I can catch it and throw an exception with a custom error message.

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
stettix: Oh and as discussed IRL, it would be really useful if you could add a test case for a multiple partition columns in `MetastoreSpec`, that was a bit of an omission in the previous version.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
stettix: I think it's nice to have well defined error messages that are the same irrespective of backend used. You can always include the underlying exception as a cause in the top level exception anyway.

Maybe consider doing something like this in `TableVersions`? https://github.com/guardian/table-versions/pull/11/files#diff-d0feac5d648547f94392ba148085a913R166

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
pvighi: I'll refactor this to make it more clear, it's a column of type date that is also called date. 
I think MetastoreSpec expects the partition column to be like that

stettix: Ah, makes sense!

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
pvighi: i think you can put _ at the end of the import list. 
But there's some other classes in that glue.model package that have the same name as our own model

stettix: I see. Makes sense as it is then!

stettix (approved):
Just some minor comments, great work! 👍👍

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
pvighi: Yeah I'll fix all of these, I just expected intellij "optimise imports" to do this on its own but I guess it doesn't

build.sbt
```diff
@@ -6,15 +6,15 @@ import com.lucidchart.sbt.scalafmt.ScalafmtCorePlugin.autoImport._
 ThisBuild / organization := "com.gu"
 ThisBuild / name := "table-versions"
 ThisBuild / description := "Version control for your Big Data!"
-
+lazy val scalatest = "org.scalatest" %% "scalatest" % "3.0.5"
```
stettix: Just remembered: this project has a file `Dependencies.scala` so it probably makes sense to put shared dependencies there.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
pvighi: I think it's good to have defined error messages that are the same regardless of implementation.
But if we want to be able to handle the expected errors it would probably be better if instead of throwing exceptions we returned errors in a more FP way like in an Either or a custom sealed trait.

If we stick to throwing exceptions it would be more useful if we defined different exception classes that can be caught and handled independently

stettix: We're not really _throwing_ exceptions though, we're just using `Throwable` objects as part of failed `IO`s as that's what it uses. But I guess one option is to use some monad stack e.g. `EitherT[IO, E, T]` or something. But that's a bigger discussion of course...

As for this case, it's not a big deal, but it seems odd if the tests for the base trait need to know all the variations of error messages that the implementations may produce. I guess another is to not check the returned error at all.

pvighi: yeah I agree that the test should not have to handle different cases like this.
maybe for now we can check that some error was returned and also an additional check we could make is that the version has not changed

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
TBonnin: `ProfileCredentialsProvider` is necessary for accessing aws services from your local machine.
We shouldn't have guardian specific settings here though.
We can use `ProfileCredentialsProvider()` that `Creates a new profile credentials provider that returns the AWS security credentials configured for the default profile`) and assume dev default profile is set to `ophan` in their `~/.aws/config`

https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/auth/profile/ProfileCredentialsProvider.html#ProfileCredentialsProvider--

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+  lazy val credentials = new AWSCredentialsProviderChain(
+    new EnvironmentVariableCredentialsProvider(),
+    new SystemPropertiesCredentialsProvider(),
+    new ProfileCredentialsProvider("ophan"),
+    new ProfileCredentialsProvider(),
+    new InstanceProfileCredentialsProvider(false)
+  )
+  val glue: AWSGlue = AWSGlueClient.builder().withCredentials(credentials).withRegion(region).build()
+  val schema = "temp"
+  def tableLocation(tableName: String) = new URI(s"s3://ophan-temp-schema/test-data/$tableName")
```
TBonnin: should `schema` and `tableLocation` be settings rather than hardcoded value only related to the guardian Data Lake

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,88 @@
...
+
+  val partitionedTable = {
+    val tableName = "test_partitioned_" + dedupSuffix
+
+    TableDefinition(TableName(schema, tableName),
+                    tableLocation(tableName),
+                    PartitionSchema(List(PartitionColumn("date"))))
+  }
+
+  "A metastore implementation" should behave like metastoreWithSnapshotSupport(IO { new GlueMetastore(glue) },
```
TBonnin: Any reason not to use simple parenthesis? ie: `IO(new GlueMetastore(glue))`

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,165 @@
...
+import com.gu.tableversions.metastore.{Metastore, VersionPaths}
+
+import scala.collection.JavaConversions._
+
+class GlueMetastore[F[_]](glue: AWSGlue)(implicit F: Sync[F]) extends Metastore[F] {
+
+  override def currentVersion(table: TableName): F[TableVersion] = {
+
+    def getPartitionColumns(glueTable: GlueTable): List[PartitionColumn] =
+      Option(glueTable.getPartitionKeys).toList.flatten map { glueColumn =>
```
TBonnin: I think we should avoid mixing dot notation and infix. I would propose to stick with dot notation everywhere in the codebase.

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  private def getPartitionedTableVersion(
+      table: TableName,
+      tableLocation: URI,
+      partitionColumns: List[PartitionColumn]): F[TableVersion] = {
+
+    def toPartition(columnValues: List[ColumnValue]): Partition = columnValues match {
+      case head :: tail => Partition(head, tail: _*)
+      case _            => throw new Exception("empty columnValues list for partition found")
```
TBonnin, line 46: Would it be better to have the return type of this function to explicitely mention that an exception can be thrown.
Understanding where this exception is being handled is not obvious

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
pvighi: I'll just write a recap here of our talk on friday:
When writing this I assumed Teamcity would not have the credentials to access aws but if it can be given access to an aws account it would be better. 
Otherwise we could just run it locally and maybe add a trigger so that it runs before pushing or something to like that to avoid people forgetting to run it.

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    def toPartitionWithVersion(gluePartition: GluePartition): (Partition, Version) = {
+
+      val partitionColumnAndValue: List[(PartitionColumn, String)] = partitionColumns.zip(gluePartition.getValues)
+      val columnValues: List[ColumnValue] = partitionColumnAndValue.map { case (key, value) => ColumnValue(key, value) }
```
TBonnin: nitpick: You can save a few key keystrokes here I think: `partitionColumnAndValue.map(ColumnValue.tupled)`

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
pvighi: Regarding referencing the "ophan" profile explicitly, we could extract the profile name to a configuration file or env variable or something

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -0,0 +1,165 @@
...
+      else
+        unversionedLocation + version.label
+    }
+    for {
+      glueTable <- getGlueTable(table)
+      tableLocation <- F.delay(findTableLocation(glueTable))
+      location = partitionLocation(tableLocation)
+      storageDescriptor <- F.delay(new StorageDescriptor().withLocation(location))
+      partitionValues <- F.delay(partition.columnValues.map(_.value).toList)
+      input <- F.delay(new PartitionInput().withValues(partitionValues).withStorageDescriptor(storageDescriptor))
```
TBonnin: Are the 3 statements above not pure and requiring `F.delay` rather than `F.pure`?

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
pvighi: yeah that sounds reasonable.

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
pvighi: I could just do `tableLocation = findTableLocation` if I'm not worried about exceptions right? 
I don't remember in this particular case but In most cases I added delay when I'm calling the aws java code that is out of our control so It might throw exceptions

TBonnin (commented):
Thanks @stettix that looks very good.
I have a few questions, especially about the usage of `F.delay` for functions that seems to be pure

stettix:
> Thanks @stettix that looks very good.

All credit is due to @pvighi for this one! :)

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
stettix: Yes I think it's very good to wrap calls to Java code in an effect. I just wonder if it would be worth doing it in the return type of `findTableLocation` instead of here, to do it as close to the Java code as possible, and maybe helping to ensure that one doesn't accidentally forget to do it in the caller. Minor point obviously.

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
stettix: I think this is, like equivalent cases in the `SparkHiveMetastore`, cases that should never happen, i.e. invariants. Here for example, `getPartitionedTableVersion` should only be called for a partitioned table which will produce parittions with a valid set of colum values.

It would be nice if this can be narrowed down via types, for example making `partitionColumns` a `NonEmptyList` here, changing the way this is passed down. Don't know if this is really practical here though. If not, maybe the error in the `Exception` could be made clearer to indicate that this is an internal error? I.e. it could say "Invalid state: empty columnValues list for partition found", or "Internal error: ..." or something like that.

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
stettix: I think it's reasonable to wrap any calls to Java in `F.delay`, although I guess some of these cases might not need it.

pvighi: yeah originally these were just assignments without the arrow like in like 87. Then I thought that references to the java library could potentially throw exceptions and wrapped everything in delay (maybe too much)
but I'm probably being too cautious here I'll change it back.

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
```diff
@@ -154,34 +156,43 @@ trait MetastoreSpec {
...
+              List(
+                UpdatePartitionVersion(Partition(dateCol, "2019-03-01"), version)
+              )
             )
           )
-        )
+          .redeem(
+            error => Left(error),
+            success => Right(success)
+          )
```
stettix, line 187: There is an `F.fromEither` that does this for you.

pvighi: I think fromEither converts  an `Either[Throwable, A]` into `IO[A]` what I want to do here is kind of the opposite:  convert the `IO[A]` into an `IO[Either[Throwable, A]` so that it will never stop the for comprehension.

stettix: Ah OK I misread that then. Just out of interest, why do we not want the overall computation to fail in this case?

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
pvighi: good point, we could put the AWS profile name there as well.

To be honest table locations don't really matter here and Glue doesn't seem to complain if you create tables with fake paths.
I was just trying to avoiding any unforeseen issues pointing the tables to a bucket that exists and we own but it could be anything really.

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
pvighi: not really, I just kinda built this following the example of the spark hive metastore and it was written like that there as well: 
https://github.com/guardian/table-versions/blob/master/spark/src/test/scala/com/gu/tableversions/spark/SparkHiveMetastoreSpec.scala#L26

I don't really care that much either way

metastore/src/test/scala/com/gu/tableversions/metastore/MetastoreSpec.scala
pvighi: because I added an extra assertion to check the current version hasn't changed. 
So I want the for comprehension to continue so that the current version is retrieved in the next step and the scenario returns the updateResult, initial version and the version after the attempted update

stettix: Ah gotcha, makes sense!

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
pvighi: columnValues is created combining 2 lists returned by glue: one with the partition column names and the other with the corresponding values. 
I can make the partition column names one a nonEmpty list because the code is already explicitly checking in line 32 that there is at least one element. But the column names can still technically be empty even though it would make no sense for glue to return that.
The problem here is just made explicit by our definition of partition that requires a nonEmptyList so this error must be handled in some way.
In other parts of this repo we just threw exceptions in situations like this but we might want to refactor and make everything return proper types with room for errors if we think it's worth it.

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
stettix: I'm not sure what the preferred style is, eve the Cats effect tutorial uses both `IO {...}` and `IO ( ... )`! https://typelevel.org/cats-effect/tutorial/tutorial.html

glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
stettix: Good points. We probably should improve the testing of error scenarios like this and fix up the code to match. Not in this PR though!

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+
+import scala.util.{Properties, Random}
+
+class GlueMetastoreSpec extends FlatSpec with Matchers with BeforeAndAfterAll with MetastoreSpec {
+
+  def readMandatoryEnvVariable(varName: String) =
+    Properties.envOrNone(varName).toRight(s"$varName environment variable must be set")
+
+  val AWSProfileEnvVarName = "TABLE_VERSIONS_TEST_AWS_PROFILE"
+  val AWSRegionEnvVarName = "TABLE_VERSIONS_TEST_AWS_REGION"
```
pvighi, line 24: I think it should be possible to just load the aws_region from your aws profile configuration but it doesn't seem to work so I had to explicitly add it as another variable here. Maybe I'm doing something wrong

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,115 @@
...
+
+  val envVars = for {
+    schema <- readMandatoryEnvVariable(SchemaEnvVarName)
+    awsRegion <- readMandatoryEnvVariable(AWSRegionEnvVarName)
+    awsProfile = Properties.envOrNone(AWSProfileEnvVarName)
+  } yield (schema, awsRegion, awsProfile)
+
+  envVars match {
+    case Left(error)                                 => cancel(error)
+    case Right((schema, awsRegion, maybeAwsProfile)) => runWithVariables(schema, awsRegion, maybeAwsProfile)
```
pvighi, line 38: I would normally make the dev params the default values if the env variables are not found, but I got the sense we want to remove all guardian specific references from the code so the test is aborted if the variables are not there.

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
TBonnin: No big deal then :P

PR #13, opened 2019-04-29
Implement writing of Spark datasets via a proxy filesystem

gr211:
This PR changes the way we write Spark datasets to use a proxy filesystem that translates paths between versioned paths and the logical partition paths that Spark deals with. This means that we can transparently use Spark functions to write partitioned data, instead of implementing our own dataset write code as previously.

See the README in this PR for details.

Note about the PR: We've spent some time to update the original PR to more thoroughly test it, trying make this reliable and production ready. Hence there are a large number of commits on top of the initial proof-of-concept implementations.


metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
```diff
@@ -31,8 +32,17 @@ object VersionPaths {
...
     else {
       def normalised(path: String): String = if (path.endsWith("/")) path else path + "/"
+
       def versioned(path: String): String = s"$path${version.label}"
+
       new URI(versioned(normalised(partitionPath.toString)))
     }
 
+  def flattenMap(partitions: List[Partition]): Map[String, String] = {
+    import scala.language.postfixOps
```
stettix: We tend not to use this in other parts of this codebase so I'd go with plain `.map` here as well.

spark/src/test/scala/org/apache/hadoop/fs/versioned/S3TestSuite.scala
```diff
@@ -0,0 +1,53 @@
+package org.apache.hadoop.fs.versioned
+
+import com.amazonaws.auth.{AWSStaticCredentialsProvider, AnonymousAWSCredentials}
+import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration
+import com.amazonaws.services.s3.AmazonS3ClientBuilder
+import com.gu.tableversions.spark.SparkHiveSuite
+import io.findify.s3mock.S3Mock
+import org.scalatest.{BeforeAndAfterAll, Suite}
+
+trait S3TestSuite extends BeforeAndAfterAll {
```
stettix: Do we still need this? I see it's mixed in to a test but doesn't seem used?

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
```diff
@@ -31,8 +32,17 @@ object VersionPaths {
       partitionPath
     else {
       def normalised(path: String): String = if (path.endsWith("/")) path else path + "/"
+
       def versioned(path: String): String = s"$path${version.label}"
+
       new URI(versioned(normalised(partitionPath.toString)))
     }
 
+  def flattenMap(partitions: List[Partition]): Map[String, String] = {
```
stettix: I think this produces Hadoop configuration entries. Maybe a more descriptive name and/or a comment would be useful here?

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+    import scala.collection.JavaConverters._
+
+    mappings = conf.iterator.asScala
+      .filter(_.getKey.startsWith(VersionPaths.KEY_PREFIX))
+      .map(_.getValue)
+      .toList
+      .map { v =>
+        (v, v + "/" + version)
+      }
+      .toMap
```
stettix: Would probably make sense to put this in `VersionPaths` alongside the function that produces these config elements? They seem to logically belong together, and you could add a round-trip test as well then.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+
+  private var baseURI: URI = _
+  private var baseFS: FileSystem = _
+  private var mappings: Map[String, String] = _
+  private var version: String = _
+
+  override def initialize(name: URI, conf: Configuration): Unit = {
+    val cacheDisabled = conf.getBoolean("fs.versioned.impl.disable.cache", false)
+    val baseFSScheme = conf.get("fs.versioned.baseFS")
+    version = conf.get("fs.versioned.version")
```
stettix: This seems to use the same version number for all partitions?

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
stettix: See comment on the filesystem proxy, but I would expect the version for each partition to go in here as well.

Update: thinking about it, the code to convert to and from Hadoop config isn't really related to the other code in this file, so I think it should move into file with the `VersionedFilesystem` code, or its own separate file.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+
+    require(Objects.nonNull(baseFSScheme), "fs.versioned.baseFS not set in configuration")
+    require(Objects.nonNull(version), "fs.versioned.version not set in configuration")
+    require(cacheDisabled, "fs.versioned.impl.disable.cache not set to true in configuration")
```
stettix: Nice check! 👍

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.hadoop.fs._
+import org.apache.hadoop.fs.permission.FsPermission
+import org.apache.hadoop.util.Progressable
+
+class VersionedFileSystem extends FileSystem {
+
+  private var baseURI: URI = _
+  private var baseFS: FileSystem = _
+  private var mappings: Map[String, String] = _
+  private var version: String = _
```
stettix: I think we'd want to make these thread safe.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+      case (acc, _) => acc
+    }
+
+    new Path(new URI(s"${uri.getScheme}:$sp"))
+  }
+
+  private lazy val setScheme: String => Path => Path =
+    scheme => path => new Path(new URI(s"$scheme:${path.toUri.getSchemeSpecificPart}"))
+
+  private lazy val toSchemeAndVersion: String => Path => Path = scheme => setScheme(scheme) andThen appendVersion
```
stettix: Any particular reason to use this function syntax here? And why are these values `lazy`?

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.hadoop.util.Progressable
+
+class VersionedFileSystem extends FileSystem {
+
+  private var baseURI: URI = _
+  private var baseFS: FileSystem = _
+  private var mappings: Map[String, String] = _
+  private var version: String = _
+
+  override def initialize(name: URI, conf: Configuration): Unit = {
```
stettix: It would be useful to have some logging in this method to be able to debug the way that a versioned filesystem is initialised.

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
```diff
@@ -8,6 +8,7 @@ import com.gu.tableversions.core.{Partition, Version}
   * Encodes the mapping between version numbers and storage paths.
   */
 object VersionPaths {
+  val KEY_PREFIX = "versioned__"
```
stettix: Wouldn't it be more standard practice to prefix config keys with something like `com.gu.tableversions` or similar?

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -117,24 +118,21 @@ object VersionedDataset {
...
+    VersionPaths.flattenMap(partitionPaths.keys.toList) foreach {
+      case (k, v) => spark.sparkContext.hadoopConfiguration.set(k, v)
     }
+
+    val partitions = table.partitionSchema.columns.map(_.name)
+
+    dataset.write
+      .mode(SaveMode.Append)
+      .partitionBy(partitions: _*)
+      .parquet(VersionedFileSystem.SCHEME + "://" + table.location.getPath)
```
stettix: Probably worth leaving in the comment about that points out that the format needs to be customisable?

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+  private lazy val setScheme: String => Path => Path =
+    scheme => path => new Path(new URI(s"$scheme:${path.toUri.getSchemeSpecificPart}"))
+
+  private lazy val toSchemeAndVersion: String => Path => Path = scheme => setScheme(scheme) andThen appendVersion
+
+  private lazy val baseVersionedPath: Path => Path = path => toSchemeAndVersion(baseURI.getScheme)(path)
+  private lazy val versionedPath: Path => Path = path => setScheme(VersionedFileSystem.SCHEME)(path)
+
+  override protected def checkPath(path: Path): Unit = super.checkPath(baseVersionedPath(path))
+}
```
stettix: There are quite a few methods on `Filesystem` that aren't overridden here, that probably should be. If you have a look at this example proxy: https://github.com/apache/hive/blob/master/shims/common/src/main/java/org/apache/hadoop/fs/ProxyFileSystem.java, you'll see it overrides a number of other methods.

I think we want to ensure that the versioning layer is completely water tight in that it can not leak the underlying filesystem to the client in any scenario. So I think it would be worth going through the other methods and ensuring they're all handled (i.e. path translated both ways).

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+    val fileStatus = baseFS.getFileStatus(setScheme(baseURI.getScheme)(f))
+
+    new FileStatus(
+      fileStatus.getLen,
+      fileStatus.isDirectory,
+      fileStatus.getReplication.toInt,
+      fileStatus.getBlockSize,
+      fileStatus.getModificationTime,
+      versionedPath(fileStatus.getPath)
+    )
```
stettix: There are a few places where `FileStatus` objects are converted so it'd be worth pulling this out into a separate method. See for example the `swizzleFileStatus` method here: https://github.com/apache/hive/blob/master/shims/common/src/main/java/org/apache/hadoop/fs/ProxyFileSystem.java#L61 (but don't copy the name! 😁 )

spark/src/test/scala/org/apache/hadoop/fs/versioned/S3TestSuite.scala
stettix: Actually, I found the place where it's used. I don't think it adds much value over just using local files as the underlying filesystem?

spark/src/test/scala/org/apache/hadoop/fs/versioned/VersionedFileSystemSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+class VersionedFileSystemSpec
+    extends FlatSpec
+    with Matchers
+    with SparkHiveSuite
+    with S3TestSuite
+    with S3SparkTestSuite {
+
+  spark.sparkContext.setLogLevel("ERROR")
+
+  "flattening List[Partition]" should "produce configuration entries" in {
```
stettix: This test should go to a spec for wherever the Hadoop conversion functions end up.

spark/src/test/scala/org/apache/hadoop/fs/versioned/VersionedFileSystemSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+import com.gu.tableversions.spark.SparkHiveSuite
+import org.apache.spark.sql.SaveMode
+import org.scalatest.{FlatSpec, Matchers}
+
+class VersionedFileSystemSpec
+    extends FlatSpec
+    with Matchers
+    with SparkHiveSuite
+    with S3TestSuite
+    with S3SparkTestSuite {
```
stettix: I think we should have a think about how to unit test the `VersionedFilesystem` class. The test below that writes via Spark is useful but I think some direct unit tests for the class itself would be valuable too.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  private lazy val appendVersion: Path => Path = path => {
```
stettix: Maybe add some unit tests for this function?

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
gr211: This method (name quite possibly unintuitive aside) is turning `Partition`s into strings like `part1=value1/part2=value2`. It does not per se add it to spark's configuration itself (`VersionedDataset.writeVersionedPartitions()` does that)

I'll still give it a better name for now.

stettix:
Might be worth having a quick look at this: https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/filesystem/testing.html#Testing_a_new_filesystem
To see if we can use some of these existing tests to check our implementation.

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
stettix: I know it doesn't _add_ to the configuration, but it is producing properties for the hadoop config, no? So it seems logically to belong with the details of the `VersionedFilesystem` code.

gr211: Actually, I am sort of wrong. It does have a `KEY_PREFIX` and sits already half way through to preparing stuff for spark's config

pvighi: I agree, the name of this def is probably not descriptive enough.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: Yes in this current impl

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: Don't think it's needed. AFAICT, spark never reruns `initialize` on a previously existing object, and our own methods do not modify them

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: `lazy` is possibly a tad of an premature optimisation. The reason was this FS is not cached, so until any of those methods is called, it'd create the said FS faster.

As for the style, I quite like this haskell'ish syntax

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: It would be nice, however with a non-cached FS, that will litter the output. We'll have to try

stettix: Good point.

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
gr211: ouch, indeed!

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,132 @@
...
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs._
+import org.apache.hadoop.fs.permission.FsPermission
+import org.apache.hadoop.util.Progressable
+
+class VersionedFileSystem extends FileSystem {
+
+  private var baseURI: URI = _
+  private var baseFS: FileSystem = _
+  private var mappings: Map[String, String] = _
```
pvighi: minor: 
maybe this could have a name that describes from what to what it maps ?
something like BaseToVersionedPaths or something like that

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
stettix: I doubt very much the time taken to initialise a `val` that points to a function is noticeable compared to loading a filesystem object and writing a Spark dataset through it. It might even be slower (also in a very insignificant way of course), as `lazy` incurs a synchronisation overhead...

As for Haskell syntax, the rest of the code base doesn't use it so I'd rather change it.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
```diff
@@ -0,0 +1,132 @@
...
+      fileStatus.getBlockSize,
+      fileStatus.getModificationTime,
+      versionedPath(fileStatus.getPath)
+    )
+  }
+
+  private lazy val appendVersion: Path => Path = path => {
+    val uri = path.toUri
+
+    val sp = mappings.foldLeft(uri.getSchemeSpecificPart) {
```
pvighi: what does sp stand for? can we rename it to something more clear?

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
stettix: Strictly speaking, a value that's assigned to these `var`s in `initialize` may not be visible to other threads reading it later. It may be unlikely to be a problem, but I'd rather be 100% sure. Just adding `@volatile` annotations should be enough to fix it.

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -117,24 +119,24 @@ object VersionedDataset {
...
-        datasetForPartition.write
-          .parquet(partitionPath.toString) // TODO: Take in format parameter. Or a dataset writer?
+    VersionedFileSystem.partitionMappings(partitionPaths.keys.toList) foreach {
+      case (k, v) => spark.sparkContext.hadoopConfiguration.set(k, v)
     }
+
+    val partitions = table.partitionSchema.columns.map(_.name)
+
+    dataset.write
+      .mode(SaveMode.Append)
```
stettix: I don't think we want `Append` mode here, the default behaviour of writing but blowing up if files already exist at the given location seems better as it might catch errors? Not a biggie as it shouldn't make a difference when everything works well!

examples/src/test/scala/com/gu/tableversions/examples/MultiPartitionTableLoaderSpec.scala
```diff
@@ -25,6 +27,8 @@ class MultiPartitionTableLoaderSpec extends FlatSpec with Matchers with SparkHiv
     implicit val tableVersions: TableVersions[IO] = InMemoryTableVersions[IO].unsafeRunSync()
     implicit val metastore: Metastore[IO] = new SparkHiveMetastore[IO]()
 
+    spark.sparkContext.hadoopConfiguration.set("fs.versioned.baseFS", "file")
```
stettix: Maybe set this in the `TableLoader` code so you don't have to repeat it in each test.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: I find this style encourages function composition (which it does here on a very small scale). If that's not FP... ;-) 

I guess i'll go with the consensus as i don't mind too much either way, just trying to be more FP and break away from old java habits :). @pvighi @desbo, what's your preference ?

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: Good point. Done

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
desbo: I tend to avoid a having too strong of a personal preference in favour of consistency with the rest of the codebase... but as it happens, I'd write these as `def`s anyway.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: I have added the extra methods that `ProxyFileSystem` does have. Although it does not break the tests, it works fine without them. For now, I am leaving them in as comments for ease of testing once against a read s3 filesystem.

I did try to run the tests against a local s3 back-end, direct with S3AFS and through this proxy, and get all sorts of errors around renaming files, permissions etc. In other words, adding those methods did not help. The pb is somewhere else (and is probably just test plumbing) but a real test run will sort this one way or the other.

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
gr211: Changed

spark/src/test/scala/org/apache/hadoop/fs/versioned/S3TestSuite.scala
gr211: I removed it

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
gr211: Renamed and moved to FS's companion object

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: I don't really like coding defensively as there might be nothing to fix here.

I checked a few of the other spark FS and they all have plenty of private vars being assigned upon the `initialize` call. If spark was exhibiting race conditions due to the lack of explicit memory barriers, it most likely would show in the coding style of those other FS too.

I was also reading things like https://stackoverflow.com/questions/1850270/memory-effects-of-synchronization-in-java which dwelves around code similar to that of `FileSystem` where `fs` is created out of `synchronized` blocks but should be flushed alongside the `CACHE` variable as the later refers to the former.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: Changed to defs. I still would find 
```
val baseVersionedPath: Path => Path = path => (setScheme(scheme) andThen appendVersion)(path)
```
easier to read than 
```
def baseVersionedPath(path: Path): Path = appendVersion(setScheme(baseURI.getScheme, path))
```

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
gr211: I stuffed it (for now at least) into VersionedFileSystem's companion object.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: Moved the other method out of `VersionPaths` (as per other comment below). Might obsolete this change. 
Also, this would need revisiting if/when the proxy fs is changed to accomodate a different version for each partition.

spark/src/test/scala/com/gu/tableversions/spark/VersionedDatasetSpec.scala
```diff
@@ -52,37 +55,54 @@ class VersionedDatasetSpec extends FlatSpec with Matchers with SparkHiveSuite {
...
+    VersionedDataset.writeVersionedPartitions(eventsGroup2, table, version2, partitionPaths)
 
     // Check that data was written to the right place.
 
-    readDataset[Event](tableUri.resolve(s"date=2019-01-15/${version.label}"))
-      .collect() should contain theSameElementsAs List(
-      Event("101", "A", Date.valueOf("2019-01-15")),
-      Event("102", "B", Date.valueOf("2019-01-15"))
-    )
+    val path = tableUri.resolve(s"table/").toString.replace("file:", "versioned://")
```
stettix: Or maybe just `replace("file:", "versioned:")`?

build.sbt
```diff
@@ -50,12 +50,15 @@ lazy val cli = project
...
-      "log4j" % "log4j" % "1.2.17"
-    ) ++ sparkDependencies)
+  .settings(libraryDependencies ++= Seq(
+    "commons-io" % "commons-io" % "2.6",
+    "com.typesafe.scala-logging" %% "scala-logging" % "3.9.0",
+    "log4j" % "log4j" % "1.2.17",
+    "com.amazonaws" % "aws-java-sdk" % "1.11.542" % Test,
+    "org.apache.hadoop" % "hadoop-common" % "2.9.2" % Test,
+    "org.apache.hadoop" % "hadoop-aws" % "2.9.2" % Test,
+    "io.findify" %% "s3mock" % "0.2.5" % Test
```
stettix: We can remove the `s3mock` dependency now. Also, I don't think we need `aws-java-sdk` or `hadoop-aws` any more either?

examples/src/test/scala/com/gu/tableversions/examples/DatePartitionedTableLoaderSpec.scala
```diff
@@ -42,6 +44,8 @@ class DatePartitionedTableLoaderSpec extends FlatSpec with Matchers with SparkHi
     implicit val tableVersions: TableVersions[IO] = InMemoryTableVersions[IO].unsafeRunSync()
     implicit val metastore: Metastore[IO] = new SparkHiveMetastore[IO]()
 
+    spark.sparkContext.hadoopConfiguration.set("fs.versioned.baseFS", "file")
+
```
stettix, line 48: I think we should encapsulate this in a function in `VersionedFilesystem`, e.g. `setUnderlyingScheme` or something. I.e. I don't think the client code should have to fiddle with hadoop config directly.

spark/src/main/scala/com/gu/tableversions/spark/VersionedFileSystem.scala
```diff
@@ -0,0 +1,159 @@
...
+    import scala.collection.JavaConverters._
+
+    val partitionMappings = conf.iterator.asScala
+      .filter(_.getKey.startsWith(VersionedFileSystem.keyPrefix))
+      .map(_.getValue)
+      .toList
+      .map { v =>
+        (v, v + "/" + version)
+      }
+      .toMap
```
stettix: It should be good to factor out this code into a separate method, then unit testing that. (Or maybe not the bit that interacts with the Hadoop config, but everything else?)

spark/src/test/scala/com/gu/tableversions/spark/SparkHiveSuite.scala
```diff
@@ -77,6 +77,8 @@ trait SparkHiveSuite extends BeforeAndAfterAll with BeforeAndAfterEach with Lazy
       .set("spark.driver.host", "127.0.0.1")
       .set("spark.sql.orc.impl", "native")
       .set("spark.sql.shuffle.partitions", "8") // For speeding up tests. The default is 200, see https://spark.apache.org/docs/latest/sql-programming-guide.html#other-configuration-options
+      .set("fs.versioned.impl", "com.gu.tableversions.spark.VersionedFileSystem")
+      .set("fs.versioned.impl.disable.cache", "true")
```
stettix: I don't think this should be set here as this isn't used in production code. Would be good to encapsulate these two steps in a method, e.g. `VersionedFileSystem.initialise()`. Then call it from `VersionedDataset.insert...` perhaps?

spark/src/test/scala/com/gu/tableversions/spark/VersionedDatasetSpec.scala
```diff
@@ -130,8 +150,64 @@ class VersionedDatasetSpec extends FlatSpec with Matchers with SparkHiveSuite {
     tableUpdate.userId shouldBe userId
   }
 
+  "Inserting multiple records into the same partition" should "write the correct data to the filesystem" in {
+    spark.sparkContext.hadoopConfiguration.set("fs.versioned.baseFS", "file")
```
stettix: Could replace these calls with a single one outside of each test? (Perhaps replacing it with a method call if we add one for this...)

spark/src/test/scala/com/gu/tableversions/spark/VersionedFileSystemSpec.scala
```diff
@@ -0,0 +1,33 @@
...
+//
+//    val fs = new VersionedFileSystem
+//    fs.initialize(new URI(s"file:///tmp/test"), ???)
+//
+//    val path = new Path(new URI(s"file:///tmp/test"))
+//
+////    VersionedFileSystem.partitionMappings(m) shouldBe Map(
+////      "versioned__date=2019-01-01/hour=01" -> "date=2019-01-01/hour=01")
+//
+//  }
```
stettix: I guess we should delete this for now, until we decide how to further unit test the proxy filesystem.

desbo:
Just a heads up that I'm going to work on the outstanding comments here. I've marked a bunch of older ones as resolved to help me see what needs doing.

spark/src/main/scala/com/gu/tableversions/spark/PathMapper.scala
```diff
@@ -0,0 +1,76 @@
...
+    val pathWithoutVersionDirectory = partitionMappings
+      .find {
+        case (_, versionedPartition) =>
+          path.toString.contains(versionedPartition)
+      }
+      .map {
+        case (partition, versionedPartition) =>
+          new Path(path.toString.replace(versionedPartition, partition))
+      }
+      .getOrElse(setVersionedScheme(path))
```
stettix: Given we set the scheme at the end of this method, I guess we just need `.getOrElse(path)` here?

examples/src/test/scala/com/gu/tableversions/examples/MultiPartitionTableLoaderSpec.scala
```diff
@@ -17,6 +17,8 @@ import org.scalatest.{FlatSpec, Matchers}
   */
 class MultiPartitionTableLoaderSpec extends FlatSpec with Matchers with SparkHiveSuite {
 
+  spark.sparkContext.setLogLevel("ERROR")
+
```
stettix, line 22: Might have been mentioned elsewhere, but we could move these to the `TableLoader` class used by all these specs.

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
```diff
@@ -31,8 +31,9 @@ object VersionPaths {
       partitionPath
     else {
       def normalised(path: String): String = if (path.endsWith("/")) path else path + "/"
+
       def versioned(path: String): String = s"$path${version.label}"
+
```
stettix: I don't disagree with the change, but this one is actually causing a conflict with the master branch, so it might just be easier to leave this as it is for now...

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
desbo: Not knowing too much about Java concurrency, is there a downside to adding `@volatile`? It might not strictly be necessary, but it at least guarantees (as far as I can tell) that these vars would be available to all threads after assignment. It's true that other `FileSystem` implementations mutate variables in `initialise` without the annotation, but if it doesn't have an overhead I'd be tempted to add it defensively.

stettix: The only downside I can see is a tiny overhead in the added memory barrier, but I would imagine that's on many orders of magnitude smaller than the actual I/O operations these methods represent. It probably doesn't make a difference in practice but I would tend to prefer correctness over performance by default.

stettix:
I spent some time double checking which methods of `Filesystem` that the proxy should overwrite. I came up with this list of methods to override in addition to the ones we have already:

```
  override def setReplication(f: Path, replication: Short): Boolean
  override def setTimes(f: Path, mtime: Long, atime: Long): Unit
  override def setOwner(f: Path, username: String, groupname: String): Unit
  override def deleteOnExit(f: Path): Boolean
  override def cancelDeleteOnExit(Path f): Boolean  // Not in the proxy in Hive, but if we do the other on we should do this one too!
  override def getFileChecksum(f: Path): FileChecksum
```

I don't think any of them are likely to make a difference in the scenario we're using now (Spark + S3) but I think we might as well add these for completeness.

The remaining methods either delegate to other methods that we have already overridden or don't involve the underlying filesystem at all.

spark/src/main/scala/com/gu/tableversions/spark/VersionedFileSystem.scala
```diff
@@ -0,0 +1,159 @@
...
+  override def append(f: Path, bufferSize: Int, progress: Progressable): FSDataOutputStream =
+    baseFS.append(pathMapper.forUnderlying(f), bufferSize, progress)
+
+  override def rename(src: Path, dst: Path): Boolean =
+    baseFS.rename(pathMapper.forUnderlying(src), pathMapper.forUnderlying(dst))
+
+  override def delete(f: Path, recursive: Boolean): Boolean =
+    baseFS.delete(pathMapper.forUnderlying(f), recursive)
+
+  private def toggleFileStatus(fileStatus: FileStatus, isParam: Boolean): FileStatus = {
```
stettix: How about replacing the `isParam` argument with an argument `convert: Path => Path` instead? Then the caller can just pass in `pathMapper.fromUnderlying` or `pathMapper.forUnderlying`.

(And change the name from `toggle...` to `convert...`  or `map...`?)

stettix:
Actually, thinking about it, I don't think we need to override `deleteOnExit`  and `cancelDeleteOnExit` as the `processDeleteOnExit` method which uses these values delegates to `delete` which is already handled.

spark/src/main/scala/org/apache/hadoop/fs/versioned/VersionedFileSystem.scala
gr211: We should refrain from coding defensively. And favor consistency.

> [...] the rest of the code base doesn't use it so I'd rather change it. @stettix 

> I tend to avoid a having too strong of a personal preference in favour of consistency with the rest of the codebase... @desbo 

in this very same PR.

> I would tend to prefer correctness over performance by default

Can it be proven to behave incorrectly without such annotations (in the context of spark, not some theoretical contrived test with a 1000 threads accessing a non-`synchronized` spark cache map)?

If not, it's just syntactic noise at best.

stettix: Consistency with what? Are there other cases where we mutate `var`s that are accessible across threads?

gr211: WIth other spark's filesystem implementations. That seems relevant to me.

stettix: Hadoop code has a different set of practices, same for Spark code. As mentioned above, I prefer going with what's correct rather than what probably works and that other bits of code does. (Otherwise we'd be writing this filesystem implementation in Java for a start...)

I would also say that I don't think it's worth spending much time on debating this point.

gr211: Indeed.

stettix:
> Might be worth having a quick look at this: https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/filesystem/testing.html#Testing_a_new_filesystem
> To see if we can use some of these existing tests to check our implementation.

I had a look at it, it was messy and I decided it wasn't worthwhile.

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -37,6 +37,27 @@ object Partition {
...
+      case ColumnValueRegex(columnName, value) => Right(ColumnValue(PartitionColumn(columnName), value))
+      case _                                   => Left(new Exception(s"Invalid partition string: $partitionStr"))
+    }
+
+    partitionStr.split("/").toList.traverse(parseColumnValue).flatMap {
+      case head :: tail => Right(Partition(head, tail: _*))
+      case _            => Left(new Exception(s"Empty partition string found: $partitionStr"))
+    }
+  }
+
```
stettix, line 70: This code moved here from the `SparkHiveMetastore` class.

metastore/src/main/scala/com/gu/tableversions/metastore/VersionPaths.scala
```diff
@@ -11,20 +11,6 @@ import com.gu.tableversions.core.{Partition, Version}
...
-      partitions: List[Partition],
-      version: Version,
-      tableLocation: URI): Map[Partition, URI] = {
-
-    partitions.map { partition =>
-      val partitionBasePath = partition.resolvePath(tableLocation)
-      partition -> pathFor(partitionBasePath, version)
-    }.toMap
-  }
-
```
stettix, line 27: This, and associated tests, are no longer needed.

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -115,24 +116,19 @@ object VersionedDataset {
...
-          .parquet(partitionPath.toString) // TODO: Take in format parameter. Or a dataset writer?
-    }
-  }
+    val partitions = table.partitionSchema.columns.map(_.name)
 
+    dataset.write
+      .partitionBy(partitions: _*)
+      .mode(SaveMode.Append)
+      .parquet(VersionedFileSystem.scheme + "://" + table.location.getPath) // TODO: Take in format parameter. Or a dataset writer?
+  }
```
stettix, line 133: This is really the key change of this PR: instead of writing partitions manually, we get Spark to do it for us, the versioning of partitions being handled by the underlying filesystem proxy.

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -52,10 +53,10 @@ object VersionedDataset {
         datasetPartitions <- IO(VersionedDataset.partitionValues(dataset, table.partitionSchema)(dataset.sparkSession))
 
         // Resolve the path that each partition should be written to, based on their version
```
TBonnin: Is this comment still relevant?

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -37,6 +42,32 @@ object Partition {
 
   case class ColumnValue(column: PartitionColumn, value: String)
 
+  private val ColumnValueRegex =
+    """(?x)
+      |([a-z][a-z0-9_]*)  # column name
```
gr211, line 47: cant remember whether java/scala supports non-greedy, but if it does,  but maybe change `*` to `*?`, to avoid `a=1/b=2` being 2 extra matches

spark/src/main/scala/com/gu/tableversions/spark/filesystem/ProxyFileSystem.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  // A utility to map between Paths for the proxy and underlying FileSystems.
+  @volatile protected var pathMapper: PathMapper = _
+
+  /** Call from concrete implementation to initialise the base class. */
+  protected def initialiseProxyFileSystem(_baseUri: URI, _pathMapper: PathMapper, hadoopConfig: Configuration): Unit = {
+    baseUri = _baseUri
+    pathMapper = _pathMapper
+
+    baseFs = FileSystem.get(baseUri, hadoopConfig)
```
tomwadeson: Wondering whether it might be preferable to pass in a configured and initialised `baseFs: FileSystem` rather than doing that here?

spark/src/main/scala/com/gu/tableversions/spark/filesystem/PathMapper.scala
```diff
@@ -0,0 +1,91 @@
...
+/**
+  * Path mapper that converts paths for partitions according to a pre-configured set of versioned partitions.
+  */
+class VersionedPathMapper(underlyingFsScheme: String, partitionVersions: Map[Partition, Version]) extends PathMapper {
+
+  import VersionedPathMapper._
+
+  override def forUnderlying(path: Path): Path = {
+    assert(path.toUri.getScheme == VersionedFileSystem.scheme,
+           s"Path provided to `forUnderlying` ($path) not in the ${VersionedFileSystem.scheme} scheme")
```
TBonnin: Can we avoid runtime check in favor of compile time check? ie: `Path` type is used to represent two different things. Can they be two different types so one cannot construct the wrong type of Path?

spark/src/main/scala/com/gu/tableversions/spark/filesystem/PathMapper.scala
```diff
@@ -0,0 +1,91 @@
...
+      }
+      .map {
+        case (_, version) =>
+          new Path(path.toString.replace(s"/${version.label}", ""))
+      }
+      .getOrElse(path)
+  }
+
+  private def appendVersion(path: Path): Path = {
+    val uri = path.toUri
```
gr211: inline it ?

spark/src/main/scala/com/gu/tableversions/spark/filesystem/PathMapper.scala
```diff
@@ -0,0 +1,91 @@
...
+    val uriSchemeSpecificPart = uri.getSchemeSpecificPart
+
+    val versionForPath = partitionVersions
```
gr211: Minor point. `normalize(partition.toString)` is called 3 times in a very short space. Maybe `map` to `case (k, v) => (normalize(k.toString), v)` or similar upfront

spark/src/main/scala/com/gu/tableversions/spark/filesystem/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+Note that to provide these configuration settings via Spark config, you have to prefix the config keys with "spark.hadoop.".
+The method `VersionedFilesystem.sparkConfig` is a convenience method that will create the right configuration parameters for you.
+
+## Implementation notes
+
+### Configurating the underlying file system
+
+Currently, we only support selecting the underlying file system schema via a global setting.
+This means that a job using this code can only write to a single type of underlying storage.
```
TBonnin, line 39: What would be a use case where we would like to write to multiple type of storages?

spark/src/main/scala/com/gu/tableversions/spark/filesystem/ProxyFileSystem.scala
```diff
@@ -0,0 +1,103 @@
...
+
+  override def create(
+      f: Path,
+      permission: FsPermission,
+      overwrite: Boolean,
+      bufferSize: Int,
+      replication: Short,
+      blockSize: Long,
+      progress: Progressable): FSDataOutputStream =
+    baseFs.create(pathMapper.forUnderlying(f), permission, overwrite, bufferSize, replication, blockSize, progress)
```
gr211, line 54: Minor. The `forUnderlying` and `toUnderlying` would read a tad nicer if they pimped `Path`, ie `f.forUnderlying` ?

spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
stettix: Nope, I'll delete it!

spark/src/main/scala/com/gu/tableversions/spark/filesystem/PathMapper.scala
stettix: I'm not sure we can in any useful way. The path mapper converts the paths that are given to it by the `ProxyFilesystem`. If we made this API more strongly typed, the filesystem would have to wrap the `Path`s it gets from the calling code (in Spark) to the right type. But I don't think this would be less error prone than making sure to call the right method in the first place.

spark/src/main/scala/com/gu/tableversions/spark/filesystem/README.md
stettix: Well, one can imagine a job that writes one table to hdfs and another to s3. Or, one that writes newer partitions to s3 or older ones to s3 (to keep recent data on a fast filesystem).

As I mention though, not something we need at the moment.

core/src/main/scala/com/gu/tableversions/core/model.scala
stettix: Interesting. This regex is only used to each part of a string that has already been split by `/` so I think we're OK here, might not be worth changing?

spark/src/main/scala/com/gu/tableversions/spark/filesystem/PathMapper.scala
```diff
@@ -0,0 +1,91 @@
...
+import java.net.URI
+
+import com.gu.tableversions.core.{Partition, Version}
+import org.apache.hadoop.fs.Path
+
+/**
+  * Map paths between one "outer" FileSystem (e.g. a ProxyFileSystem) and an underlying one.
+  * For example, by changing the schema, or appending version directories.
+  */
+trait PathMapper {
```
tomwadeson: I think it'd be good to locate this at `ProxyFileSystem.PathMapper`

tomwadeson (commented):
This looks really tidy. A couple of comments/suggestions, but neither's important. Nice work all!

spark/src/main/scala/com/gu/tableversions/spark/filesystem/PathMapper.scala
stettix: Agreed.

core/src/main/scala/com/gu/tableversions/core/model.scala
gr211: Probably good enough then since already split around `/` indeed

spark/src/main/scala/com/gu/tableversions/spark/filesystem/ProxyFileSystem.scala
stettix: I had a look at this but it's a bit messy as the path mapper would have to be an implicit argument then, and it's a `var` which can't be implicit...

As this is just used internally in this class I think it's reasonable to leave it as it is.

spark/src/main/scala/com/gu/tableversions/spark/filesystem/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+...and data will be written to the right path for each partition.
+
+To use this file system in a job, you need to configure this file system. You do this via the following Hadoop configuration properties:
+
+```
+fs.versioned.impl=com.gu.tableversions.spark.VersionedFileSystem
+fs.versioned.impl.disable.cache=true
+fs.versioned.baseFS=<the scheme of the underlying filesystem, e.g. "s3" or "hdfs"
+fs.versioned.configDirectory=<the path where the partition version configuration will be written>
```
desbo, line 28: Maybe we should create a helper method on `VersionedFileSystem` to set all of these.

desbo: ...I've just seen you've already done that. Good stuff!

desbo (approved):
LGTM

PR #18, opened 2019-05-03
PartitionColumn fix for kryo

gr211:
PartitionColumn serialisation fails with

```
java.lang.ClassCastException: com.gu.tableversions.core.Partition$PartitionColumn cannot be cast to java.lang.String
```

### DIFF
--- a/core/src/main/scala/com/gu/tableversions/core/model.scala
+++ b/core/src/main/scala/com/gu/tableversions/core/model.scala
@@ -30,7 +30,7 @@ object Partition {
   /** Convenience constructor for multiple partitions. */
   def apply(first: ColumnValue, rest: ColumnValue*): Partition = Partition(NonEmptyList(first, rest.toList))
 
-  case class PartitionColumn(name: String) extends AnyVal
+  case class PartitionColumn(name: String)
 
   case class ColumnValue(column: PartitionColumn, value: String)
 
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/com/gu/tableversions/spark/VersionedDataset.scala
```diff
@@ -32,14 +32,12 @@ object VersionedDataset {
       * @return a tuple containing the updated table version information, and a list of the changes that were applied
       *         to the metastore.
       */
-    def versionedInsertInto(table: TableDefinition, userId: UserId, message: String)(
-        implicit tableVersions: TableVersions[IO],
-        metastore: Metastore[IO]): (TableVersion, TableChanges) =
+    def versionedInsertInto(table: TableDefinition, userId: UserId, message: String): (TableVersion, TableChanges) =
```
stettix, line 35: 👍

core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -30,7 +30,10 @@ object Partition {
   /** Convenience constructor for multiple partitions. */
   def apply(first: ColumnValue, rest: ColumnValue*): Partition = Partition(NonEmptyList(first, rest.toList))
 
-  case class PartitionColumn(name: String) extends AnyVal
+  /**
+    * Note: do not add `extends AnyVal`. It breaks kryo serialisation.
+    */
+  case class PartitionColumn(name: String)
```
stettix, line 36: 👍

stettix (approved):
Nice fix! 👍

PR #19, opened 2019-05-03
remove unused vals from glue metastore test

pvighi:
just realised I left some unused vals on my last refactor of the glue metastore test


stettix (approved):
Deleting code is good! 👍

PR #20, opened 2019-05-09
Group all versioning concerns under a `VersionContext` umbrella

tomwadeson:
Having a play with application wiring and dependency passing -- see what you think.

### DIFF
--- a/examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
+++ b/examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
@@ -2,10 +2,8 @@ package com.gu.tableversions.examples
 
 import java.time.Instant
 
-import cats.effect.IO
 import com.gu.tableversions.core.TableVersions.{UpdateMessage, UserId}
-import com.gu.tableversions.core.{TableDefinition, TableVersions, Version}
-import com.gu.tableversions.metastore.Metastore
+import com.gu.tableversions.core.{TableDefinition, TableVersions}
 import com.gu.tableversions.spark.{SparkSupport, VersionContext}
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.spark.sql.{Dataset, SparkSession}
--- a/examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
+++ b/examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
@@ -6,7 +6,7 @@ import cats.effect.IO
 import com.gu.tableversions.core.TableVersions.{UpdateMessage, UserId}
 import com.gu.tableversions.core.{TableDefinition, TableVersions, Version}
 import com.gu.tableversions.metastore.Metastore
-import com.gu.tableversions.spark.VersionedDataset
+import com.gu.tableversions.spark.VersionContext
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.spark.sql.{Dataset, SparkSession}
 
@@ -15,14 +15,16 @@ import scala.reflect.runtime.universe.TypeTag
 /**
   * Example code for loading data into versioned tables and updating the versions of such tables.
   */
-class TableLoader[T <: Product: TypeTag](table: TableDefinition, createTableDdl: String, isSnapshot: Boolean)(
-    implicit tableVersions: TableVersions[IO],
-    metastore: Metastore[IO],
-    generateVersion: IO[Version],
-    spark: SparkSession)
+class TableLoader[T <: Product: TypeTag](
+    versionContext: VersionContext,
+    table: TableDefinition,
+    createTableDdl: String,
+    isSnapshot: Boolean)(implicit spark: SparkSession)
     extends LazyLogging {
 
   import spark.implicits._
+  import versionContext._
+  import versionContext.implicits._
 
   def initTable(userId: UserId, message: UpdateMessage): Unit = {
     // Create table in underlying metastore
@@ -36,8 +38,6 @@ class TableLoader[T <: Product: TypeTag](table: TableDefinition, createTableDdl:
     spark.table(table.name.fullyQualifiedName).as[T]
 
   def insert(dataset: Dataset[T], userId: UserId, message: String): Unit = {
-    import VersionedDataset._
-
     val (latestVersion, metastoreChanges) = dataset.versionedInsertInto(table, userId, message)
 
     logger.info(s"Updated table $table, new version details:\n$latestVersion")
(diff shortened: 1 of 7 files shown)


examples/src/test/scala/com/gu/tableversions/examples/DatePartitionedTableLoaderSpec.scala
```diff
@@ -170,3 +170,13 @@ object DatePartitionedTableLoaderSpec {
...
+
+object TestVersionContext {
+
+  def default(implicit spark: SparkSession): IO[VersionContext] =
+    for {
+      tableVersions <- InMemoryTableVersions[IO]
+      metastore = new SparkHiveMetastore[IO]()
+      versionGenerator = Version.generateVersion
+    } yield VersionContext(tableVersions, metastore, versionGenerator)
+}
```
stettix: ❤️

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
```diff
@@ -17,46 +17,32 @@ import org.apache.spark.sql.{Dataset, Row, SaveMode, SparkSession}
...
-      */
-    def versionedInsertInto(table: TableDefinition, userId: UserId, message: String): (TableVersion, TableChanges) =
-      versionedInsertDatasetIntoTable(delegate, table, userId, message).unsafeRunSync()
-
-  }
+final case class VersionContext(
+    tableVersions: TableVersions[IO],
+    metastore: Metastore[IO],
+    generateVersion: IO[Version]
+) {
```
stettix: This is a great change, just one thought: I wonder if it would make sense to move this `VersionContext` class into `core`? The `versionedInsertDatasetIntoTable` would then have to move out of it again, of course.

The reason I'm wondering about this is that this could make a good place to hang some functionality that interacts with both the `TableVersions` and the `Metastore`. For example the `TableLoader.checkout()` method: https://github.com/guardian/table-versions/blob/master/examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala#L47. This currently lives in example code, but it would be useful to have this somewhere in the real code. It doesn't belong in either of the two underlying services as it interacts with both. It also doesn't rely on Spark or any of the concrete implementations, so `core` would seem like a good place for it.

Likewise, the "commit and sync metastore" parts of the current `versionedInsertDatasetIntoTable` method could perhaps be pulled into common code int he same place.

What do you think?

tomwadeson: Yep, that sounds like a better place for it. Would you like to jump in and move things around? I think you'll have a better sense for exactly where to locate things.

It would be nice to preserve the ergonomics of enriching `Dataset`/`Dataframe` with a versioned insert method *without* having to pass the `VersionContext` around implicitly. I think I've got an idea brewing for this, but would need to play with it.

stettix: OK, I can have a quick go and see what it looks like. (Looking forward to hearing about your cunning idea!)

stettix: So I had a play but without much luck. One issue is that the `Metastore` trait is in the separate `metastore` module; thinking about it, I think this module should be removed and the code go in `core`, as these are implementation-neutral interfaces and types, just as the `TableVersions` trait and related code.

Also, the `VersionContext` you've created here is really a Spark version context, which makes a lot of sense, collecting the bits needed to write Spark code in an easy to use package. That's probably a separate concept from any lower level service that doesn't know about Spark, and is useful even if such a thing is added. So I think the changes here are fine as they are.

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
```diff
@@ -87,6 +73,27 @@ object VersionedDataset {
...
+        * @return a tuple containing the updated table version information, and a list of the changes that were applied
+        *         to the metastore.
+        */
+      def versionedInsertInto(table: TableDefinition, userId: UserId, message: String): (TableVersion, TableChanges) =
+        versionedInsertDatasetIntoTable(delegate, table, userId, message).unsafeRunSync()
+    }
+  }
+}
+
+object VersionContext {
```
stettix: As a slight aside, I think there are some more `implicit`s you can get rid of in this code if you want, on the `partitionValues` and `writeVersionedPartitions` method?

examples/src/test/scala/com/gu/tableversions/examples/DatePartitionedTableLoaderSpec.scala
```diff
@@ -171,3 +171,13 @@ object DatePartitionedTableLoaderSpec {
...
```
stettix: BTW maybe pull out into its own file as it's used in a couple of tests?

examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
```diff
@@ -6,7 +6,7 @@ import cats.effect.IO
 import com.gu.tableversions.core.TableVersions.{UpdateMessage, UserId}
 import com.gu.tableversions.core.{TableDefinition, TableVersions, Version}
 import com.gu.tableversions.metastore.Metastore
-import com.gu.tableversions.spark.VersionedDataset
+import com.gu.tableversions.spark.VersionContext
```
stettix: I think you can remove some `import`s here now.

spark/src/test/scala/com/gu/tableversions/spark/VersionContextSpec.scala
```diff
@@ -105,19 +105,25 @@ class VersionedDatasetSpec extends FlatSpec with Matchers with SparkHiveSuite {
...
+    val versionContext = {
+      val stubMetastore: Metastore[IO] = new StubMetastore(
+        currentVersion = initialTableVersion,
+        computedChanges = stubbedChanges
+      )
+
+      val tableVersions: TableVersions[IO] = (for {
+        t <- InMemoryTableVersions[IO]
+        _ <- t.init(usersTable.name, isSnapshot = true, UserId("test"), UpdateMessage("init"), Instant.now())
+      } yield t).unsafeRunSync()
```
stettix, line 120: This isn't related to your changes, but this bit of code is repeated a few times. Might be nice pulling it out somewhere. Could `TestVersionContext` be used here as well maybe? It would have to move into the `spark` module to do that, of course. But maybe that would be a step too far...

examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
tomwadeson: Who needs `-Ywarn-unused:imports` with eyes like that? 😀

stettix (approved):
Nice work! 👍

spark/src/test/scala/com/gu/tableversions/spark/VersionContextSpec.scala
tomwadeson: I did have a look at this, but nothing jumped out quickly.

There's a bit of me that quite likes it anyway - makes it easy to see the behaviour that we're going to observe when interacting with the `VersionContext` in the test.

Will keep an eye on it!

PR #21, opened 2019-05-10
Add support for arbitrary file formats for tables

stettix:
This adds a 'file format' field to table definitions, so that tables can be written in any format that Spark supports.

### DIFF
--- a/core/src/main/scala/com/gu/tableversions/core/model.scala
+++ b/core/src/main/scala/com/gu/tableversions/core/model.scala
@@ -90,7 +90,7 @@ final case class TableName(schema: String, name: String) {
   def fullyQualifiedName: String = s"$schema.$name"
 }
 
-final case class TableDefinition(name: TableName, location: URI, partitionSchema: PartitionSchema) {
+final case class TableDefinition(name: TableName, location: URI, partitionSchema: PartitionSchema, format: FileFormat) {
   def isSnapshot: Boolean = partitionSchema == PartitionSchema.snapshot
 }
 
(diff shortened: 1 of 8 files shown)


core/src/main/scala/com/gu/tableversions/core/model.scala
```diff
@@ -100,3 +100,10 @@ final case class TableDefinition(name: TableName, location: URI, partitionSchema
 sealed trait TableVersion
 final case class PartitionedTableVersion(partitionVersions: Map[Partition, Version]) extends TableVersion
 final case class SnapshotTableVersion(version: Version) extends TableVersion
+
+case class FileFormat(name: String) extends AnyVal
```
tomwadeson, line 104: Do you think there'd be any value in locking this down to a closed set of known-supported formats?

tomwadeson (approved):
👍

core/src/main/scala/com/gu/tableversions/core/model.scala
JonNorman, line 104: Yeah I guess to Tom's point, is there a reason we don't want to use enumeratum here?

stettix: Not really, I don't think it makes sense for this library to make assumptions as to which set of formats that client code will be using.

stettix: As mentioned above, I think this library should be agnostic about what file formats client code uses. If our data lake jobs are restricted to a limited set, then I think that policy should be enforced in ophan-data-lake, not here.

stettix: Oh and just to add to that, the set of formats supported by Spark is open-ended (you can provide a custom format), so I don't think it makes much sense to try to enumerate all known Spark file formats here either.

tomwadeson: TIL!

PR #23, opened 2019-05-13
add format information to glue metastore

pvighi:
Fixed a glue with the metastore where the serialization library, input format and output format are left blank when creating partitions. This caused the table not to work properly.

To fix this the metastore will now copy the current table settings every time it performs any operation that involves a glue storageDescriptor (adding partitions or creating a new table or partition version)

### DIFF
--- a/glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
+++ b/glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
@@ -1,6 +1,7 @@
 package com.gu.tableversions.glue
 
 import java.net.URI
+
 import cats.effect.Sync
 import cats.implicits._
 import com.amazonaws.services.glue.AWSGlue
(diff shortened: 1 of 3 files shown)


glue/src/main/scala/com/gu/tableversions/glue/GlueMetastore.scala
```diff
@@ -112,18 +115,22 @@ class GlueMetastore[F[_]](glue: AWSGlue)(implicit F: Sync[F]) extends Metastore[
...
+      table: TableName,
+      partition: Partition,
+      version: Version): F[StorageDescriptor] =
     for {
       gluetable <- getGlueTable(table)
       tableLocation = findTableLocation(gluetable)
       partitionLocation = partition.resolvePath(tableLocation)
       versionedPartitionLocation = VersionPaths.pathFor(partitionLocation, version)
-    } yield versionedPartitionLocation
+    } yield extractFormatParams(gluetable.getStorageDescriptor).withLocation(versionedPartitionLocation.toString)
```
pvighi, line 130: it would probably be more technically correct to get the existing partition storage descriptor and keep the existing parameters while updating just the location. But that would be an extra api call and the partitions are created by cloning the table storage descriptor so the end result should be the same unless the partition is created or updated bypassing this GlueMetastore

glue/src/test/scala/com/gu/tableversions/glue/GlueMetastoreSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+  it should "ignore empty input or output format without failing" in {
+    val serdeWithNoLibrary = new SerDeInfo().withSerializationLibrary(testSerialisationLib)
+    val descriptorWithNoFormats = new StorageDescriptor().withSerdeInfo(serdeWithNoLibrary)
+    val actual = GlueMetastore.extractFormatParams(descriptorWithNoFormats)
+    actual.getInputFormat should be(null)
+    actual.getOutputFormat should be(null)
+    actual.getSerdeInfo.getSerializationLibrary shouldBe testSerialisationLib
+  }
+
+}
```
stettix, line 61: Nice tests!

stettix (approved):
Looks good! 👍

PR #24, opened 2019-05-13
Metastore refactorings

stettix:
This PR contains a couple of refactorings:

* Removing the `metastore module`, moving the code in it into `core`.

The `Metastore` trait and related code isn't really any less core than `TableVersions` etc so it doesn't make much sense to leave it out of core.

* Creating a new `VersionedMetastore` which combines the `TableVersions` and `Metastore` interface and implements functionality that can be implemented purely in terms of the underlying traits, e.g. setting the metastore to refer to a specific version in history.

This refactoring is enabled by the previous step, and simplifies the application API for versioning, not having to deal with two separate APIs.

As for naming, I think this combination can be viewed as a "version enabled metastore", i.e. a metastore that can remember previous states.

This is a a "draft" until other pending metastore changes have been merged so that these can be merged into this.

### DIFF
--- a/examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
+++ b/examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
@@ -2,8 +2,8 @@ package com.gu.tableversions.examples
 
 import java.time.Instant
 
+import com.gu.tableversions.core.TableDefinition
 import com.gu.tableversions.core.TableVersions.{UpdateMessage, UserId}
-import com.gu.tableversions.core.{TableDefinition, TableVersions}
 import com.gu.tableversions.spark.{SparkSupport, VersionContext}
 import com.typesafe.scalalogging.LazyLogging
 import org.apache.spark.sql.{Dataset, SparkSession}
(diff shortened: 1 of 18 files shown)


examples/src/main/scala/com/gu/tableversions/examples/TableLoader.scala
```diff
@@ -43,15 +43,4 @@ class TableLoader[T <: Product: TypeTag](
...
-    val checkout = for {
-      _ <- tableVersions.setCurrentVersion(table.name, id)
-      newVersion <- tableVersions.currentVersion(table.name)
-      currentMetastoreVersion <- metastore.currentVersion(table.name)
-      changes = metastore.computeChanges(currentMetastoreVersion, newVersion)
-      _ <- metastore.update(table.name, changes)
-    } yield ()
-
-    checkout.unsafeRunSync()
-  }
```
stettix, line 56: This has moved into the `VersionedMetastore` class.

The remaining `TableLoader` class is now so anemic that it might be worth removing it - just using the versioning API directly is pretty much as concise. I think that's a good sign!

pvighi:
looks good to me  so far

tomwadeson (approved):
Nice change. Seems to fall out naturally from what's already there - a good sign!

stettix:
Merging this from a separate PR after rebasing on top of @pvighi 's changes in a separate PR.

PR #25, opened 2019-05-14
Metastore refactorings

stettix:
This PR contains a couple of refactorings:

* Removing the `metastore module`, moving the code in it into `core`.

The `Metastore` trait and related code isn't really any less core than `TableVersions` etc so it doesn't make much sense to leave it out of core.

* Creating a new `VersionedMetastore` which combines the `TableVersions` and `Metastore` interface and implements functionality that can be implemented purely in terms of the underlying traits, e.g. setting the metastore to refer to a specific version in history.

This refactoring is enabled by the previous step, and simplifies the application API for versioning, not having to deal with two separate APIs.

As for naming, I think this combination can be viewed as a "version enabled metastore", i.e. a metastore that can remember previous states.

This is a a "draft" until other pending metastore changes have been merged so that these can be merged into this.

### DIFF
--- a/glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreIntegrationSpec.scala
+++ b/glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreIntegrationSpec.scala
@@ -6,27 +6,17 @@ import cats.effect.IO
 import cats.implicits._
 import com.amazonaws.auth._
 import com.amazonaws.auth.profile.ProfileCredentialsProvider
-import com.amazonaws.services.glue.model.{
-  Column,
-  CreateTableRequest,
-  DeleteTableRequest,
-  GetPartitionsRequest,
-  GetTableRequest,
-  SerDeInfo,
-  StorageDescriptor,
-  TableInput
-}
+import com.amazonaws.services.glue.model._
 import com.amazonaws.services.glue.{AWSGlue, AWSGlueClient}
 import com.gu.tableversions.core.Partition.PartitionColumn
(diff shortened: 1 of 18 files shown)


stettix:
Based on draft PR https://github.com/guardian/table-versions/pull/24, but rebased on top of the changes in https://github.com/guardian/table-versions/pull/23

glue/src/it/scala/com/gu/tableversions/glue/GlueMetastoreIntegrationSpec.scala
```diff
@@ -6,27 +6,17 @@ import cats.effect.IO
...
-import com.gu.tableversions.core._
-import com.gu.tableversions.metastore.MetastoreSpec
-import org.scalatest.{Assertion, BeforeAndAfterAll, FlatSpec, Matchers}
+import com.gu.tableversions.core.{MetastoreSpec, _}
+import org.scalatest.{Assertion, FlatSpec, Matchers}
 
 import scala.util.{Properties, Random}
 
-class GlueMetastoreSpec extends FlatSpec with Matchers with MetastoreSpec {
+class GlueMetastoreIntegrationSpec extends FlatSpec with Matchers with MetastoreSpec {
```
pvighi, line 17: 👍 I was considering renaming this myself but then I forgot to do it

stettix: I only did it because IntelliJ complained about it...

pvighi (approved):
looks good 👍

PR #26, opened 2019-05-14
Generalise effect type of generating Version instances

stettix:
`IO` => `F` when building Version instances

### DIFF
--- a/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
+++ b/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
@@ -2,7 +2,7 @@ package com.gu.tableversions.spark
 
 import java.time.Instant
 
-import cats.effect.Effect
+import cats.effect.{Effect, Sync}
 import cats.implicits._
 import com.gu.tableversions.core.Metastore.TableChanges
 import com.gu.tableversions.core.TableVersions.TableOperation._
@@ -54,7 +54,7 @@ object SparkSupport {
       dataset: Dataset[T],
       table: TableDefinition,
       userId: UserId,
-      message: String)(implicit F: Effect[F]): F[(TableVersion, TableChanges)] = {
+      message: String)(implicit F: Sync[F]): F[(TableVersion, TableChanges)] = {
 
     import versionContext._
 
--- a/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
+++ b/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
@@ -4,10 +4,10 @@ import java.time.Instant
 
 import cats.effect.IO
 import cats.implicits._
-import com.gu.tableversions.core.Metastore.TableChanges
 import com.gu.tableversions.core.TableVersions.TableOperation._
 import com.gu.tableversions.core.TableVersions._
 import com.gu.tableversions.core._
+import com.gu.tableversions.core.Metastore.TableChanges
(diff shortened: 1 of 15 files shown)


tomwadeson (commented):
Could do `VersionContext` while you're at it?

stettix:
> Could do `VersionContext` while you're at it?

I had a go at that actually, but: The `SparkSupport` methods wants to run the effect because they throw exceptions (they don't return `IO`s, as they match conventions of Spark `Dataset` methods. So it would only be the `VersionContext` case class that is generalised to `F`, all of the rest of the code in this file would still use `IO`, so it didn't seem worthwhile.

What do you think?

tomwadeson:
I see!

A couple of options....

1) Break the Spark conventions (I don't think this is a bad option TBH)
2) Constrain `F[_]` to something more powerful than `Sync`, like `Effect`, then you can `.toIO.unsafeRunSync` it. A nice property of this is that the types tell us that you need to run these actions as side-effects to provide a familiar Spark API!

stettix:
> A couple of options....
> 
> 1. Break the Spark conventions (I don't think this is a bad option TBH)
> 2. Constrain `F[_]` to something more powerful than `Sync`, like `Effect`, then you can `.toIO.unsafeRunSync` it. A nice property of this is that the types tell us that you need to run these actions as side-effects to provide a familiar Spark API!

I like 2, so just pushed out a version doing this. See what you think!

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
```diff
@@ -37,7 +37,8 @@ final case class SparkSupport(versionContext: VersionContext) {
         *         to the metastore.
         */
       def versionedInsertInto(table: TableDefinition, userId: UserId, message: String): (TableVersion, TableChanges) =
-        SparkSupport.versionedInsertDatasetIntoTable(versionContext, delegate, table, userId, message).unsafeRunSync()
+        F.toIO(SparkSupport.versionedInsertDatasetIntoTable(versionContext, delegate, table, userId, message))
```
tomwadeson, line 40: With `import cats.effect.implicits._` you'd be able to `.toIO.unsafeRunSync`, if you preferred (I'm fine with it as it is)

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
```diff
@@ -48,31 +49,31 @@ object SparkSupport {
...
     */
-  def versionedInsertDatasetIntoTable[T](
-      versionContext: VersionContext,
+  def versionedInsertDatasetIntoTable[T, F[_]](
+      versionContext: VersionContext[F],
       dataset: Dataset[T],
       table: TableDefinition,
       userId: UserId,
-      message: String): IO[(TableVersion, TableChanges)] = {
+      message: String)(implicit F: Effect[F]): F[(TableVersion, TableChanges)] = {
```
tomwadeson: I think this could be `Sync` here as you only need to import, not run, side effects

tomwadeson (approved):
Nice!

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
stettix: Good to know!

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
stettix: Indeed!

PR #27, opened 2019-05-15
fix a couple of simple bugs

pvighi:
- baseFileSystemSchema param was ignored in VersionedFileSystem
- Only the path of the url was preserved when converting from underlying scheme to versioned://

### DIFF
--- a/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
+++ b/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
@@ -9,7 +9,7 @@ import com.gu.tableversions.core.Metastore.TableChanges
 import com.gu.tableversions.core.TableVersions.TableOperation._
 import com.gu.tableversions.core.TableVersions._
 import com.gu.tableversions.core._
-import com.gu.tableversions.spark.filesystem.{VersionedFileSystem, VersionedPathMapper}
+import com.gu.tableversions.spark.filesystem.VersionedFileSystem
 import com.gu.tableversions.spark.filesystem.VersionedFileSystem.VersionedFileSystemConfig
 import org.apache.spark.sql.{Dataset, Row, SaveMode}
 
--- a/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
+++ b/spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
@@ -1,5 +1,6 @@
 package com.gu.tableversions.spark
 
+import java.net.URI
 import java.time.Instant
 
 import cats.effect.{Effect, Sync}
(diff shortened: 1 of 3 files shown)


spark/src/test/scala/com/gu/tableversions/spark/VersionContextSpec.scala
```diff
@@ -277,6 +277,16 @@ class VersionContextSpec extends FlatSpec with Matchers with SparkHiveSuite {
...
+  "withVersionedScheme" should "preserve all URI parts" in {
+
+    val uri = new URI("s3", "bucketName", "/some/path/date=2019-01-01", "someQueryString", "someFragment")
+
+    val expectedUri =
+      new URI(VersionedFileSystem.scheme, "bucketName", "/some/path/date=2019-01-01", "someQueryString", "someFragment")
+
+    SparkSupport.setVersionedScheme(uri) shouldBe expectedUri
+  }
+
```
pvighi, line 289: It might be an overkill to have a test just for this, but the current tests didn't catch the problem because the local filesystem doesn't use all the parts of a uri

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
```diff
@@ -126,10 +127,19 @@ object SparkSupport {
...
-      .save(VersionedFileSystem.scheme + "://" + table.location.getPath)
+      .save(versionedUri.toString)
   }
+
+  private[spark] def setVersionedScheme(underlyingUri: URI): URI =
+    new URI(VersionedFileSystem.scheme,
+            underlyingUri.getAuthority,
+            underlyingUri.getPath,
+            underlyingUri.getQuery,
+            underlyingUri.getFragment)
```
stettix, line 144: Good fix! I think at some point it would be worth moving all our URI-manipulating code into a single place and test it thoroughly, as we do similar things in various places and it's proved to be error prone. This is absolutely fine for the time being though!

stettix (approved):
Nice! 👍

spark/src/test/scala/com/gu/tableversions/spark/VersionContextSpec.scala
stettix: It's definitely not overkill! :) In fact, if we add some common code for manipulating URIs, I'd like to test the heck out of it using property based tests... as you say, it's proved a source of errors before.

spark/src/main/scala/com/gu/tableversions/spark/VersionContext.scala
pvighi: yeah that is probably the way to go.
alternatively in some other guardian projects we use a [scala uri implementation](https://github.com/lemonlabsuk/scala-uri) that might give us what we need if we are ok with introducing that dependency

stettix: Interesting, I just looked at that library yesterday actually! It'd be worth considering, though it pulls in a couple of further dependencies (parboiled and shapeless) and may be overkill for what we need (changing schemas and rewriting paths). Something to ponder if we carry on with this I guess.

